PR #3, opened 2018-10-17
Fix CI status badge

jborrey:


jborrey:
It's now working somehow ...

PR #4, opened 2018-10-17
implement new and better npm audit scanner

as3richa:
tl;dr the old implementation spewed out the entire output of npm audit --json to Salus's STDOUT. This wasn't great, because it was basically impossible to parse out what the specific breakage or remediation is from that output.

This PR adds a fairly heavy wrapper interface around npm audit, which distills it down to the most important details - the advisory id, the module name, a link to the advisory, and indicators for whether it's in prod and whether there exists an exception in the config already. For QoL reasons it also flags unnecessary exceptions and exceptions against dev-only advisories.

See the screenshot and the new fixtures for what things actually look like

screen shot 2018-10-17 at 2 11 14 pm


lib/salus/npm_audit_wrapper.rb
```diff
@@ -0,0 +1,239 @@
...
+
+    # Some common, easily-abbreviated terms to shorten advisory titles
+    TITLE_SUBSTITUTIONS = {
+      'Regular Expression'    => 'Regex',
+      'Denial of Service'     => 'DoS',
+      'Remote Code Execution' => 'RCE'
+    }.freeze
+
+    # Cap title and module name length to avoid the table getting too wide
+    MAX_MODULE_LENGTH = 16
```
as3richa: These restrictions are roughly what it takes to prevent salus from wrapping the table. I think we should considering loosening salus's wrapping just a bit, but @jborrey's call :)

lib/salus/npm_audit_wrapper.rb
```diff
@@ -0,0 +1,239 @@
...
+      @colors = colors
+    end
+
+    def run!
+      Dir.chdir(@path) do
+        ensure_package_lock! do
+          raw = run_command!('npm audit --json')
+          raw_advisories = JSON.parse(raw, symbolize_names: true).fetch(:advisories).values
+
+          advisories = raw_advisories.map do |radv|
```
jborrey: nit: prefer to just use `advisory` for that var name.

lib/salus/npm_audit_wrapper.rb
```diff
@@ -0,0 +1,239 @@
...
+            # If the advisory exists in a prod dependency, there'll be some finding
+            # where dev is false
```
jborrey: I don't understand what is happening here. Maybe put an example of the hash of one advisory in a comment so it's obvious what this block is doing.

lib/salus/npm_audit_wrapper.rb
```diff
@@ -0,0 +1,239 @@
...
+              excepted
+            )
+          end
+
+          # Sort advisories with un-excepted prod vulnerabilities first,
+          # and in ascending order of ID otherwise
+          advisories.sort_by! do |advisory|
+            unexcepted_and_prod = !advisory.excepted? && advisory.prod?
+            [(unexcepted_and_prod ? 0 : 1), advisory.id.to_i]
+          end
```
jborrey: nice!

lib/salus/npm_audit_wrapper.rb
```diff
@@ -0,0 +1,239 @@
...
+
+          advisories_by_id = advisories.map(&:id).zip(advisories).to_h
+
+          # The build fails iff there exist un-expect advisories against prod deps
+          failing_advisory_ids = advisories.select { |adv| !adv.excepted? && adv.prod? }.map(&:id)
+
+          # For informational purposes, we record any exceptions that don't refer
+          # to an actual advisory
+          extraneous_exceptions = @exceptions.reject { |id| advisories_by_id.key?(id) }
```
jborrey: nit: slightly cleaner might be the difference of 2 sets

```ruby
extraneous_exceptions = @exceptions - advisories_by_id.keys
```

lib/salus/npm_audit_wrapper.rb
```diff
@@ -0,0 +1,239 @@
...
+          end
+
+          if advisories.any?
+            log(tabulate_advisories(advisories))
+          else
+            log(colorize('There are no advisories against your dependencies. Hooray!', :green))
+          end
+
+          if failing_advisory_ids.any?
+            s = failing_advisory_ids.join(', ')
```
jborrey: nit: prefer non one char var names

lib/salus/npm_audit_wrapper.rb
```diff
@@ -0,0 +1,239 @@
...
+
+    def log(string)
+      @stream.puts(string)
+    end
+
+    def ensure_package_lock!
+      created_package_lock = false
+
+      if !File.exist?('package-lock.json')
+        log('Creating a temporary package-lock.json...')
```
jborrey: this should only be logged if verbose option (which can be given to Salus via a `-v` flag).
Might want to do this in a separate PR to keep this one small.

lib/salus/npm_audit_wrapper.rb
```diff
@@ -0,0 +1,239 @@
...
+        run_command!('npm install --package-lock-only', raise_on_failure: true)
+        created_package_lock = true
+      end
+
+      yield
+    ensure
+      if created_package_lock
+        File.delete("package-lock.json")
+        log('Removed temporary package-lock.json')
```
jborrey: this should only be logged if verbose option (which can be given to Salus via a `-v` flag).
Might want to do this in a separate PR to keep this one small.

lib/salus/npm_audit_wrapper.rb
```diff
@@ -0,0 +1,239 @@
...
+      end
+    end
+
+    def run_command!(command, raise_on_failure: false)
```
jborrey: There is already a `run_shell()` in salus that you can use (part of the `Salus::Scanners::Base`).

jborrey: I would move all this code into the `NPMAudit` module since it will only be used there.

lib/salus/npm_audit_wrapper.rb
```diff
@@ -0,0 +1,239 @@
...
+
+      if raise_on_failure && !status.success?
+        raise "`#{command}` failed with exit code #{status.exitstatus}"
+      end
+
+      stdout
+    end
+
+    def abbreviate_title(title)
+      TITLE_SUBSTITUTIONS.each { |from, to| title = title.gsub(from, to) }
```
jborrey: would this also work?

```ruby
def abbreviate_title(title)
  title = TITLE_SUBSTITUTIONS[title].nil? ? title : TITLE_SUBSTITUTIONS[title]
  title = title[0...(MAX_TITLE_LENGTH - 1)] + '~' if title.length > MAX_TITLE_LENGTH
  title
end
```

jborrey: if so, should then be O(1) running time

jborrey:
Very slick!

I suggested moving this module just to be directly in the `NPMAudit` class. If you do this, you will then also be able to use the [`report_info()`](https://github.com/coinbase/salus/blob/master/lib/salus/scanners/base.rb#L46) method that come from the scanner base class. This is useful to inject CVE exceptions into the reports that are sent over HTTP (and eventually make it into the logging pipeline) correctly capture the state of the repo.

There is an example of that [here](https://github.com/coinbase/salus/blob/master/lib/salus/scanners/bundle_audit.rb#L26).

as3richa:
> I suggested moving this module just to be directly in the NPMAudit class

I'm fine with doing this if that's your preference, but just to explain my thought process here 😁:

Most of the scanners (brakeman, bundle audit, the old nsp one, etc.) call some arbitrary binary with no particular coupling to salus itself, and emit the result of that call pretty much verbatim. Also obviously those binaries have a life/utility outside of salus.

I tried to separate concerns similarly here: `NpmAuditWrapper` is just an arbitrary program that produces some output; salus calls it and just emits the result. The calling convention is different (a ruby class rather than a binary on disk), but I think it's the same sort of encapsulation.

lib/salus/npm_audit_wrapper.rb
jborrey: ohhh I see. okay

jborrey:
In that case, I think the separation is good, but the data around ignored advisories should be bubbled up to the report.

So perhaps we can keep `NpmAudtiWrapper` as its own unit and have it return a hash containing the text to print to STDOUT and the ignored advisories in an array, which we can then pump into the report.

To DRY up the `run_shell()` command, maybe we put that method into a module and then include it in the wrapper you have created and the base scanner.

lib/salus/scanners.rb
```diff
@@ -1,8 +1,7 @@
 module Salus::Scanners; end
 
-Dir.entries('lib/salus/scanners').each do |scanner_file|
-  next if ['.', '..'].include?(scanner_file) # don't include FS pointers
-  next unless /\.rb\z/.match?(scanner_file)  # only include ruby files
-
-  require "salus/scanners/#{scanner_file}"
+Dir.entries(File.expand_path('scanners', __dir__)).each do |filename|
```
jborrey: mm definitely, better - thanks!

bin/npm-audit-wrapper
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env ruby
+
+# This is just a stub binary for dev/debug purposes
```
jborrey: Please add a comment under here showing how you might invoke this from the command line.

lib/salus/scanners.rb
jborrey: definitely better, thanks!

lib/salus/scanners/npm_audit.rb
```diff
@@ -1,68 +1,287 @@
...
+      end
+    end
+
+    def should_run?
+      @repository.package_json_present? ||
+        @repository.package_lock_json_present? ||
+        @repository.yarn_lock_present?
+    end
+
+    def sanity_check_exceptions
```
jborrey: this and everything below can be `private`

lib/salus/scanners/npm_audit.rb
```diff
@@ -1,68 +1,287 @@
...
+
+          if json.key?(:error)
+            code = json[:error][:code] || '<none>'
+            summary = json[:error][:summary] || '<none>'
+
+            message =
+              "`#{AUDIT_COMMAND}` failed unexpectedly (error code #{code}):\n" \
+              "```\n#{summary}\n```"
+
+            log(colorize(message, :red))
```
jborrey: use [`report_error()`](https://github.com/coinbase/salus/blob/master/lib/salus/scanners/base.rb#L64) here - this will ensure that the Salus report shows that this particular scanner threw an error instead of completing successfully.

lib/salus/scanners/npm_audit.rb
```diff
@@ -1,68 +1,287 @@
...
+          advisories = raw_advisories.map do |advisory|
+            # findings is an array of objects looking roughly like
+            # [
+            #   { "version": "1.0.5", ..., "dev": false },
+            #   { "version": "1.0.5", ..., "dev": true }
+            # ]
+            # where each element records an instance of the vulnerable package
+            # in the dependency tree. If, for some finding, dev is false,
+            # then there exists a vulnerable version of the module in the prod
+            # dependency tree
```
jborrey: great comment!

lib/salus/scanners/npm_audit.rb
```diff
@@ -1,68 +1,287 @@
...
+          extraneous_exceptions = @exceptions - advisories_by_id.keys
+
+          # Also, we record any exceptions that apply only to dev dependencies
+          extraneous_dev_exceptions = @exceptions.select do |id|
+            adv = advisories_by_id[id]
+            !adv.nil? && !adv.prod?
+          end
+
+          if advisories.any?
+            log(tabulate_advisories(advisories))
```
jborrey: need to put `report_failure` above this line.
I'd then loop through and log each advisory similar to how we do here: https://github.com/coinbase/salus/blob/master/lib/salus/scanners/bundle_audit.rb#L74

I'll write a comment on the PR in a bit regarding the `log()` code.

lib/salus/scanners/npm_audit.rb
```diff
@@ -1,68 +1,287 @@
...
+          else
+            log(colorize('There are no advisories against your dependencies. Hooray!', :green))
```
jborrey: needs a `report_success` above this line.
Can remove `log()`

lib/salus/scanners/npm_audit.rb
```diff
@@ -1,68 +1,287 @@
...
+        @config.fetch('exceptions', [])
+          .map { |ex| ex['advisory_id']&.to_s }
+          .compact
+
+      @stream = StringIO.new
+
+      okay = run_audit!
+      report_stdout(@stream.string)
+
+      if okay
```
jborrey: nit: ternary operator is cleaner

lib/salus/scanners/npm_audit.rb
```diff
@@ -1,68 +1,287 @@
...
+            !adv.nil? && !adv.prod?
+          end
+
+          if advisories.any?
+            log(tabulate_advisories(advisories))
+          else
+            log(colorize('There are no advisories against your dependencies. Hooray!', :green))
+          end
+
+          if failing_advisory_ids.any?
```
jborrey: I'd then loop through and log each advisory similar to how we do here: https://github.com/coinbase/salus/blob/master/lib/salus/scanners/bundle_audit.rb#L74

I'll write a comment on the PR in a bit regarding the log() code.

lib/salus/scanners/npm_audit.rb
```diff
@@ -1,68 +1,287 @@
...
+            stringified_ids = failing_advisory_ids.join(', ')
+            log("Audit failed pending the following advisory(s): #{stringified_ids}.")
+            log('To fix the build, please resolve the previous advisory(s), or add exceptions.')
+          end
+
+          if extraneous_exceptions.any? || extraneous_dev_exceptions.any?
```
jborrey: Can also be added to the report via the `report_info()` method. Probably with a different key though, something like `extraneous-cve-exception` will do.

jborrey:
I think it will be better if we split this out into 2 PRs.

For this PR, keep all the reporting functionality you have (all the `report_info()`/`report_stdout()`), add the `report_failure()` as commented, and remove all the logging/tabulation code.

I'd make a 2nd PR that adds that stuff to the `Report` object so that we can use it for all scanners. This will also ensure that the flags like `--quite` and `--verbose` are honored.

as3richa:
I'll pull the useful stuff out of this in a bit

PR #9, opened 2018-10-25
[big refactor] implement and use a per-scan ScanReport class

as3richa:
Rather than stick everything in a big top-level hash that lives in Report, let's use a ScanReport class to record the results of a single scan. I think this gives us a better separation of concerns overall (a scanner exists independent of the entire suite) and should make it a bit easier to improve the logging story going forward.

A Report collects ScanReports (via `#add_scan_report`); `Report#to_s` calls `ScanReport#to_s` (likewise for `to_h`). `to_h` is (basically) backwards-compatible with the old version.

Some other miscellaneous changes:
- Before, calling `report_error` from a scanner would append the error to the global Salus error hash - it was associated with the scanner only by name, and not in any sort of structural way. Now, ScanReports have their own error hash and `report_error` routes errors there
- In the case that a scanner raises (and fails to handle) an exception, that error is appended both to the ScanReport's error hash and to the global error hash (because it's probably a bug)
- Before, per-scan info data were necessarily arrays, and calling `info` multiple times from a scanner with the same key would append to the same array. Now they're just scalars (`info(key, value) => @info[key] = value`)

Most of this diff is specs 😁

### DIFF
--- a/lib/salus/config.rb
+++ b/lib/salus/config.rb
@@ -48,8 +48,8 @@ module Salus
       @active_scanners   = all_none_some(SCANNERS.keys, final_config['active_scanners'])
       @enforced_scanners = all_none_some(SCANNERS.keys, final_config['enforced_scanners'])
       @scanner_configs   = final_config['scanner_configs'] || {}
-      @project_name      = final_config['project_name'].to_s || ''
-      @custom_info       = final_config['custom_info'].to_s || ''
+      @project_name      = final_config['project_name']&.to_s
+      @custom_info       = final_config['custom_info']&.to_s
       @report_uris       = final_config['reports'] || []
     end
 
@@ -59,6 +59,17 @@ module Salus
 
     def scanner_enforced?(scanner_class)
       @enforced_scanners.include?(scanner_class.to_s)
+    end
+
+    def to_h
+      {
+        active_scanners: @active_scanners.to_a.sort,
+        enforced_scanners: @enforced_scanners.to_a.sort,
+        scanner_configs: @scanner_configs,
+        project_name: @project_name,
+        custom_info: @custom_info,
+        report_uris: @report_uris
+      }.compact
     end
 
     private
--- a/lib/salus/report.rb
+++ b/lib/salus/report.rb
@@ -70,10 +70,6 @@ module Salus
       @errors[error_origin] << error_data
     end
 
-    def has_failure?(source)
-      @scans.key?(source) && @scans[source].key?("passed") && @scans[source]["passed"] == false
-    end
-
     def configuration_source(source)
       @configuration['sources'] ||= []
       @configuration['sources'] << source
--- a/lib/salus/report.rb
+++ b/lib/salus/report.rb
@@ -54,11 +54,6 @@ module Salus
       @scans[scanner][log_type] = log
     end
 
-    def salus_info(type, message)
-      @info[type] ||= []
-      @info[type] << message
-    end
-
     def salus_runtime_error(error_data)
       salus_error('Salus', error_data)
     end
--- a/lib/salus/report.rb
+++ b/lib/salus/report.rb
@@ -5,14 +5,8 @@ module Salus
   class Report
     class ExportReportError < StandardError; end
 
-    NUM_CHAR_IN_BAR = 20
-    BAR = ('=' * NUM_CHAR_IN_BAR).freeze
-    SPECIAL_BAR = ('#' * NUM_CHAR_IN_BAR).freeze
-
-    SCAN_RESULT_WORD = {
-      true => 'passed',
-      false => 'failed'
-    }.freeze
+    # FIXME(as3richa): make wrapping behaviour configurable
+    WRAP = 100
 
     CONTENT_TYPE_FOR_FORMAT = {
       'json' => 'application/json',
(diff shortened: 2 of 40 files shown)


jborrey:
Since we are changing the format of the report, I think we should [bump the version](https://github.com/coinbase/salus/blob/master/lib/salus.rb#L20).

lib/salus/config.rb
```diff
@@ -61,6 +61,17 @@ def scanner_enforced?(scanner_class)
       @enforced_scanners.include?(scanner_class.to_s)
     end
 
+    def to_h
```
jborrey: much nicer!

lib/salus/report.rb
```diff
@@ -5,160 +5,109 @@ module Salus
...
-          if verbose
-            scan_data['info']&.each do |type, messages|
-              lines << "INFO - #{type}"
-              lines += messages.map { |message| "\t#{message}" }
-            end
-          end
-        end
+    def to_s(verbose: false, wrap: WRAP)
+      output = "==== Salus Scan v#{VERSION}"
+      output += " for #{@project_name}" unless @project_name.nil?
```
jborrey: mmm that's a nice change!

lib/salus/report.rb
```diff
@@ -5,160 +5,109 @@ module Salus
...
+
+      # Sort scan reports required before optional, failed before passed,
```
jborrey: great comment and idea

lib/salus/report.rb
```diff
@@ -5,160 +5,109 @@ module Salus
...
+
+      # Only add config if verbose mode is on.
       if verbose
-        lines << "\n"
-        lines << "#{BAR} Salus Configuration #{BAR}"
-        lines += @configuration.map { |type, value| "#{type}: #{value}" }
+        # Dump config in particular as YAML rather than JSON, because salus
+        # config files are YAML. Also, stringify the keys before serializing,
+        # because the YAML module is sensitive to symbols vs strings. It's
+        # annoying
```
jborrey: haha

lib/salus/report.rb
```diff
@@ -5,160 +5,109 @@ module Salus
...
+        output += required ? ' (required)' : ' (not required)'
+      end
+
+      scan_reports.each do |report, _required|
+        output += "\n\n#{report.to_s(verbose: verbose, wrap: wrap)}"
       end
 
-      # Only add configuration if verbose mode is on.
+      # Adjust the wrap by 2 because we're wrapping indented paragraphs
+      wrap = (wrap.nil? ? nil : wrap - 2)
```
jborrey: nit: make this a new var instead of overwriting. Might stop a bug in the future if someone messes with this method.

lib/salus/report.rb
```diff
@@ -187,9 +136,44 @@ def export_report
...
     private
 
-    # Wrap lines to 100 chars by default
-    def wrap(text, line_width = 100)
-      text.gsub(/(.{1,#{line_width}})/, "\\1\n")
+    def deep_stringify_keys(datum)
+      case datum
+      when Hash then datum.map { |key, value| [key.to_s, deep_stringify_keys(value)] }.to_h
+      when Array then datum.map { |value| deep_stringify_keys(value) }
+      else datum
```
jborrey: do you need a `.to_s` on this?

lib/salus/report.rb
```diff
@@ -187,9 +136,44 @@ def export_report
 
     private
 
-    # Wrap lines to 100 chars by default
-    def wrap(text, line_width = 100)
-      text.gsub(/(.{1,#{line_width}})/, "\\1\n")
+    def deep_stringify_keys(datum)
```
jborrey: Can we just use the native ruby method? https://apidock.com/rails/Hash/deep_stringify_keys ?

lib/salus/report.rb
```diff
@@ -187,9 +136,44 @@ def export_report
...
+      case datum
+      when Hash then datum.map { |key, value| [key.to_s, deep_stringify_keys(value)] }.to_h
+      when Array then datum.map { |value| deep_stringify_keys(value) }
+      else datum
+      end
+    end
+
+    # FIXME(as3richa): factor this out along with the identical version
+    # in scan_report.rb. Maybe have a Formatting mixin module for wrapping,
```
jborrey: recommend the mixim module option

lib/salus/report.rb
```diff
@@ -187,9 +136,44 @@ def export_report
...
+          index += wrap
+        end
+      end
+
+      parts.join
+    end
+
+    def indent(text)
+      # each_line("\n") rather than split("\n") because the latter
+      # discards trailing empty lines. Also, don't indent empty lines
```
jborrey: excellent comments

lib/salus/report.rb
```diff
@@ -187,9 +136,44 @@ def export_report
...
+      text.each_line("\n").map { |line| line == "\n" ? "\n" : ("  " + line) }.join
```
jborrey: would make the string `"  "` a constant at the top or make the constant `INDENT_SIZE = 2` and then to `' ' * INDENT_SIZE` to get the string.

nit: also should use single quotes for the indent string/space.

lib/salus/scan_report.rb
```diff
@@ -0,0 +1,133 @@
...
+    end
+
+    def indent(text)
+      # each_line("\n") rather than split("\n") because the latter
+      # discards trailing empty lines. Also, don't indent empty lines
+      text.each_line("\n").map { |line| line == "\n" ? "\n" : ('  ' + line) }.join
+    end
+
+    def monotime
+      Process.clock_gettime(Process::CLOCK_MONOTONIC)
```
jborrey: nit: put comment above about what's happening here

lib/salus/scan_report.rb
```diff
@@ -0,0 +1,133 @@
...
+    end
+
+    def record
+      started_at = monotime
+
+      yield
+
+      # If the block failed to pass/fail the scan,
+      # default to pass if no errors were recorded
+      @passed = @errors.empty? if @passed.nil?
```
jborrey: This is an interesting idea.

Would this only happen if a developer wrote the scanner class incorrectly? If so I would default to failure so that this bug is more likely to be caught.

lib/salus/scan_report.rb
```diff
@@ -0,0 +1,133 @@
...
+
+    def passed?
+      @passed == true
+    end
+
+    def failed?
+      @passed == false
+    end
+
+    def info(type, value)
```
jborrey: What if you want to record multiple of the same type of info?
For instance, type might be `dependency` and there could be many values for one project.

Should we append to an array instead and make `@info` `Hash[String=>Array[String]]` ?

lib/salus/report.rb
jborrey: I see that the methods are slightly different so fine to leave this as a future TODO like you have.

lib/salus/scan_report.rb
```diff
@@ -0,0 +1,133 @@
...
+    def to_s(verbose:, wrap:)
+      banner = render_banner
+
+      # If the scan succeeded and verbose is false, just output the banner
+      # indicating pass/fail
+      return banner if @passed && !verbose
+
+      # Because we need to wrap indented paragraphs, correct the wrap
+      # by the indentation level
+      wrap = (wrap.nil? ? nil : wrap - 2)
```
jborrey: same nit as above

lib/salus/scanners/base.rb
```diff
@@ -26,55 +30,75 @@ def should_run?
...
+    #   collection and the global salus scan's error collection
+    def run!(salus_report:, required:, reraise:)
+      salus_report.add_scan_report(@report, required: required)
+
+      begin
+        @report.record { run }
+      rescue StandardError => exception
+        error_data = {
+          message: "Unhandled exception running #{name}: #{exception.class}: #{exception}",
+          klass: exception.class,
```
jborrey: nit: slight preference for this key to be `error_class`

lib/salus/scanners/base.rb
```diff
@@ -26,55 +30,75 @@ def should_run?
...
+          backtrace: exception.backtrace.take(5)
```
jborrey: noice

lib/salus/scanners/pattern_search.rb
```diff
@@ -48,55 +49,57 @@ def run
...
+                    "STDOUT: #{shell_return.stdout}\n" \
+                    "STDERR: #{shell_return.stderr}"
           end
         end
       end
 
-      if failure_messages.empty?
+      report_info(:hits, all_hits)
+      errors.each { |error| report_error('Call to sift failed', error) }
+      failure_messages.each { |message| report_error(message) }
```
jborrey: failure messages aren't really errors. They are more just what caused the scan to fail or succeed.

lib/salus/scan_report.rb
jborrey: Actually for the dep reporting scanners I see this is kind of necessary. Happy to keep then.

lib/salus/scanners/report_ruby_gems.rb
```diff
@@ -26,8 +26,11 @@ def should_run?
 
     def record_dependencies_from_gemfile_lock
       lockfile = Bundler::LockfileParser.new(@repository.gemfile_lock)
-      record_ruby_version(lockfile.ruby_version, 'Gemfile')
-      record_bundler_version(lockfile.bundler_version, 'Gemfile')
+
+      # N.B. lockfile.bundler_version is a Gem::Version
```
jborrey: what does this note mean?

spec/docker/docker_spec.rb
```diff
@@ -13,23 +13,27 @@
...
 
     Dir.chdir('spec/fixtures/docker') do
       stdout, stderr, exit_status = Open3.capture3(DOCKER_RUN_COMMAND)
     end
     expect(exit_status.success?).to eq(true), "STDOUT:\n#{stdout}\n\nSTDERR:\n#{stderr}"
-    expect(File.read(report_file_path)).to eq(
-      File.read('spec/fixtures/docker/expected_report.json').strip
-    )
+
+    # Running time is nondeterministic, so just zero it out
```
jborrey: haha good trick

spec/lib/salus/report_spec.rb
```diff
@@ -1,197 +1,176 @@
...
+  describe '#to_h and miscellaneous reporting methods' do
+    it 'emits the expected reporting data via the #to_h method' do
+      name = 'Neon Genesis Evangelion'
+      custom_info = { bitcoin_price: 100_000 }
+      config = { lemurs: 'contained', raptors: 'loose' }
+
+      report = Salus::Report.new(project_name: name, custom_info: custom_info, config: config)
+
+      # The actual content here doesn't really matter -
+      # that functionality is speced out in the ScanReport unit tests
```
jborrey: good note

spec/lib/salus/scanners/report_go_dep_spec.rb
```diff
@@ -1,83 +1,64 @@
...
+    it 'should report on all the dependencies in the Gopkg.lock file' do
+      repo = Salus::Repo.new('spec/fixtures/report_go_dep')
+      scanner = Salus::Scanners::ReportGoDep.new(repository: repo, config: {})
+
+      scanner.run
+
+      dependencies = scanner.report.to_h.fetch(:info).fetch(:dependencies)
+
+      expect(dependencies).to match_array([
+                                            {
```
jborrey: nit: rad indentation.
Would move everything to under the `expect` call

jborrey:
Exceptional PR!

This makes things much clear and robust.

lib/salus/report.rb
as3richa: nope - integers should stay integers, etc. could add a `when Symbol` case for paranoia though

lib/salus/report.rb
as3richa: That's in ActiveSupport, not Ruby 😁I'm okay with bringing in ActiveSupport if you are, but for one method i felt this was fine

lib/salus/scan_report.rb
as3richa: yeah that's exactly why it's here. i would also be okay with requiring scanners to report their status but i'm not that opinionated (and this was faster than that)

lib/salus/scanners/report_ruby_gems.rb
as3richa: `lockfile.bundler_version` isn't a string (it's a Gem::Version), which could cause some surprising breakage if you didn't stringify it first

as3richa:
> Since we are changing the format of the report, I think we should bump the version.

wasn't sure about this, because it's aaaaalmost backwards-compatible. but you're probably right. Straight to 2.0.0?

spec/lib/salus/scanners/report_go_dep_spec.rb
as3richa: oops i trusted `rubocop -a` a bit too much

lib/salus/report.rb
jborrey: Oh yeah good point. Bring it in, I'm sure we'll use it elsewhere anyway.

jborrey:
yep 2.0.0 sounds good

PR #22, opened 2018-11-12
Use Logs for Dev, Info for Reports

jborrey:
Previously, for a failed scanner, info data was presented to the
developer. This often has raw STDOUT from a scanner and other
information which makes the output that the developer needs to parse
very large and unreadable.

Instead we will use `log()` for all dev friendly output and it will be
output whenever a scanner fails. `info()` will be reserved for Salus
report consumers and verbose mode.


jborrey:
The current Salus failure is because of Yarn's bug. Fine to merge.

heimdall-asguard:
This Pull Request was merged without enough reviews
Please go [here](https://github.com/coinbase/salus/issues/23) to review and +1

PR #29, opened 2019-01-07
Integrate gosec, a static code vulnerability checker:

xolaz:
- enforced scanner
- scans any project with .go file, go.mod, go.sum, and Gopkg.lock inside of $GOPATH
- created symlink to successfully run gosec against $GOPATH projects
- has tests

gosec: https://github.com/securego/gosec


Dockerfile.tests
```diff
@@ -9,8 +9,11 @@ FROM salus-local
 
 COPY spec /home/spec
 
+# Symlink for gosec testing
+RUN ln -s /home/spec $GOPATH/src/repo/spec
+
 RUN bundle install --with test
 
-RUN rm ./spec/docker/docker_spec.rb && rm ./spec/static_analysis/rubocop_spec.rb
+RUN rm ./spec/docker/docker_spec.rb
```
xolaz: `rm ./spec/static_analysis/rubocop_spec.rb` kept failing stating file doesn't exist, I can add it back in if necessary.

jborrey: Hmmm yeah that doesn't exist anymore thanks to the new CircleCI flow. Isn't this container built in the tests? How come this was only noticed now?

lib/salus/repo.rb
```diff
@@ -13,6 +13,8 @@ class Repo
       { handle: :npmrc, filename: '.npmrc' },
       # Go
       { handle: :dep_lock, filename: 'Gopkg.lock' },
+      { handle: :go_mod, filename: 'go.mod' },
```
jborrey: Can you add comments about what these files are?
Helps for the less Go familiar folk.
I should have done the same for the others.

lib/salus/scanners/gosec.rb
```diff
@@ -0,0 +1,50 @@
...
+# https://github.com/securego/gosec
+
+module Salus::Scanners
+  class Gosec < Base
+    def run
+      # Shell Instructions:
+      #   - -fmt=json for JSON output
+      #   - gosec only successfully scans repos within $GOPATH, we force the
+      #     go path reference via symlink within Dockerfile & Dockerfile.tests
+      if ENV['RUNNING_SALUS_TESTS'] == "true"
```
jborrey: Can we avoid this branching if we add `/home/repo` to GOPATH instead of symlinking it?

spec/lib/salus/scanners/gosec_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+        expect(info[:stdout]).not_to be_nil
+        expect(info[:stdout]).not_to be_empty
+        expect(logs).to include('Potential hardcoded credentials')
+      end
+    end
+
+    context 'go project with no known vulnerabilities' do
+      let(:repo) { Salus::Repo.new('spec/fixtures/gosec/safe_goapp') }
+
+      it 'should record failure and record the STDOUT from gosec' do
```
jborrey: copy-paste error from above

jborrey:
Few comments - looking pretty good!

Dockerfile.tests
xolaz: not sure, it happened as I was building the container locally. Could the image settings be different on circleci?

lib/salus/scanners/gosec.rb
xolaz: From the digging and testing I did, it looks like we are stuck with this solution until gosec figures how to support go modules (aka until gosec stops requiring we run repo against a `$GOPATH`).  reference: https://github.com/securego/gosec/issues/234#issuecomment-441569937

lib/salus/scanners/gosec.rb
```diff
@@ -0,0 +1,50 @@
...
+        shell_return = run_shell("gosec -fmt=json /go/src/repo/#{@repository.path_to_repo}")
+      else
+        shell_return = run_shell("gosec -fmt=json /go/src/repo")
```
asla337: Would it be better to recursively scan from this directory if there are multiple folders within a project (via `/go/src/repo/...`)? If so, there might be a problem because symlinks seem to be skipped (https://github.com/securego/gosec/issues/203).

My only workaround for recursive scan was to mount a directory under `/go/src/` and specify it to salus with `-d` option. Hope this helps.

Dockerfile.tests
xolaz: I find the Dockerfile.tests really useful as well. Noting this as an improvement.

lib/salus/scanners/gosec.rb
xolaz: will try 👍

xolaz:
https://github.com/coinbase/salus/pull/36 patches failing circle tests

xolaz:
circleci integration test is failing because the result includes gosec data: 
```
      "Gosec": {
               "pass_on_raise": false
      },
```
^^ this will pass once I rebuild docker after the merge and release

xolaz:
Test cases around what scanners actually test is a great idea. It does create redundancy but I like the explicit failures it can give us depending on how scanners change as we bump them.

PR #32, opened 2019-01-15
Run unit tests inside the Salus container; run a dedicated external integration test

as3richa:
This allows us to exercise the process of building both the prod and test containers in CI. IMO it's also more useful to run the unit tests in the context of the container, because that implies that we're running against the correct version of our dependencies (Go, sift, yarn, etc). Also, this lets us replace the existing `docker_spec.rb` with an integration test of narrower scope.


.circleci/config.yml
```diff
@@ -2,38 +2,23 @@ version: 2
...
+      - run: docker run salus-tests
+      - run: bin/integration_test
   salus:
     machine: true
     steps:
       - checkout
       - run: docker run -t -v $(pwd):/home/repo coinbase/salus
   rubocop:
     docker:
-      - image: ruby:2.4.5@sha256:dd40c9f70dd028d0df8c7f7fa636daf2fc7ac1f4919562b68c60691b99714e05
```
jborrey: why not use an official ruby image?

.circleci/config.yml
```diff
@@ -2,38 +2,23 @@ version: 2
...
+      - image: circleci/ruby:2.4.5
     steps:
-      - run: apt-get update
-      - run: apt-get install -y cmake
       - checkout
-      - run: gem install bundler
-      - run: bundle install
+      - run: sudo apt-get update && sudo apt-get install cmake && bundle install
```
jborrey: if staying with the cirlce image, do we still need cmake?

bin/integration_test
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env ruby
+
+require 'open3'
+require 'json'
+
+# The incantation with which you would run salus in production
+SALUS_COMMAND = 'docker run -et --rm -v $(pwd):/home/repo salus-local'.freeze
```
jborrey: what's `-e` for?

jborrey: `-e` is for defining an envar but you dont seem to have the envar anymore

.circleci/config.yml
jborrey: let's add a sha then just to be extra cautious

PR #33, opened 2019-01-18
Bump brakeman; properly propagate brakeman errors to the user

as3richa:


as3richa:
Fine with me, but all the other fixtures under that directory are full Rails apps too :)

PR #37, opened 2019-02-01
version bump salus for release

xolaz:
version bump salus before release


xolaz:
circle tests failing due to integration tests, it will pass when I make a new build on docker

PR #40, opened 2019-02-27
Don’t attempt to run Brakeman if there is no rails app

dustMason:
Salus currently assumes that a repo is a rails app if its Gemfile contains the rails gem. Brakeman, however, exits with code 4 if it cannot find an `app` directory.

This adds a condition to the `should_run?` method of the `Brakeman` scanner subclass that causes salus to skip the Brakeman check if it sees that there is no `app` directory. This is the same method Brakeman itself uses to determine whether or not the given repo is a rails app. See https://github.com/presidentbeef/brakeman/blob/2f31fbbeb5d6aa2fce59003828b643c58496a236/lib/brakeman/scanner.rb#L25


jborrey:
Makes sense. I have no problem with this, so _ack_.

One quick test case is probably worth while (should be obvious from existing spec what to do).

I am curious to know what happened to prompt this though?

nishils:
@jborrey There's already a spec test. Do want something else? Otherwise, this looks good to me.

Basically, we have a gem that includes the rails gem but isn't a rails app on its own. @dustMason probably has more details.

jborrey:
Yeah curious to know what the use case was for including a rails gem in a not rails project.

I think we can add a test for the case where there is no `app` directory. I can make a quick commit for this.

dustMason:
It might be that this change isn't actually needed, since the project in question might be better served by including `rails` as a dev dependency in its gemspec vs. in a Gemfile. I will have time to follow up on that tomorrow.

Re: the test case, I thought that's what the newly added test case covered?

jborrey:
Oh you're right @dustMason - I completely missed that.
In any case, this might be useful for someone else so +1. @xolaz feel free to merge.

jborrey:
Why did CI stop running on this 🤔  ?

as3richa:
BTW, this broke CI. 😁

as3richa:
Oops, just noticed #42

Issue #39, opened 2019-02-23
Salus Orb


jborrey:
From #10, @raphdev brought up the following:

>A small improvement we've made is simplifying for CircleCI through an [orb](https://circleci.com/orbs/). This allows teams to integrate Salus and any configuration, simply by including the orb. The orb would have the correct docker image/version, and remote configuration locations managed by the security team. One benefit of this is that we no longer require updating several pipelines for updates that aren't in the container itself, such as handling of reporting, since these are reused from the orb.
Does something like this sound like a valuable contribution?

Thanks for bringing this to my attention! I need to keep up with the CircleCI blog more.

Yeah I certainly think that a Salus orb could be useful. Not sure if the Coinbase team wants to own the creation of that or @raphdev, are you considering making your orb general purpose to work outside of your org?

raphdev:
My intention is to make the orb general purpose. Since production CircleCI orbs are public, it makes sense to make our Salus orb useful outside our org. We'd be more than happy to contribute this :)

jborrey:
Sounds great! Thank you.

Orb source code seems public so if other orgs want to have their own version or modify it that should be easy for them to do.

jsulinski:
Is this orb publicly available? Not seeing it in the registry at https://circleci.com/orbs/registry/

Thanks!

nishils:
As far as I know, not yet. If you know how to build an orb, we could use help on that front. @raphdev are you still able to make the Salus orb public?

jsulinski:
Never have before, but gave it a shot. It's currently published as a development version.

federacy/salus@dev:0.0.1

https://github.com/federacy/salus-orb

raphdev:
We've been running a dev version in my org, but it's not public just yet. I can work on a prod release this week since I'm continuing Salus-related work for the next few weeks. 

@jsulinski Nice job! I like how you parameterized some of the Salus local config options. The orb I have doesn't expose these options, only command line flags.

Given there are two orbs in development, do we want to coordinate efforts? If so, how do we proceed?

jborrey:
Yeah nice job @jsulinski - and I appreciate the fact that you kept the default params quite strong (the README mislead me, but in the `orb.yml` is see it's different).

@raphdev I see pros and cons for both options. If we have two implementations and they are significantly different (not sure how different orbs can be ...) then it's probably worth having both. If you think they will be very similar, I'd lean towards collaboration provided that @jsulinski is happy to facilitate code reviews/merging.

[...]

PR #41, opened 2019-03-05
update readme with CI integration steps

xolaz:
docs...


jborrey:
@xolaz are you still working on this or should we close?

xolaz:
we should probably merge, I'll rebase

jborrey:
What does this add to the docs that don't already exist? Seems like just the command at the start of the README? Or am I missing something?

xolaz:
I added it based on feedback from people using Salus for the first time. The docs need general updating, but in this case they were looking for an obvious header with a what to run.

jborrey:
Okay I see. If it will be useful to others, then we should keep it.

Few ideas:
  - `coinbase/salus pulls the docker image` --> `That command will automatically pull the the Salus image published on Docker Hub.`
  - Recommend pinning the image in their configuration to a specific Salus version (and probably update the command for latest SHA). This will ensure people don't get surprises if we make breaking changes and give them a bit more defense in the case that DockerHub/CB's ability to publish new versions is compromised.
  - I remember there is that bizarre that bizarre gotcha with running Salus in CircleCI - you might want to mention using Salus's own CircleCI config as a reference.
  - When [the orb stuff gets sorted out](https://github.com/coinbase/salus/issues/39) we should add a link to this section.

heimdall-asguard:
Review Error for nishils @ 2019-06-04 04:26:11 UTC
User cannot review their own commit

heimdall-asguard:
Review Error for dnyu @ 2019-06-04 04:43:06 UTC
User must have write permissions to review

heimdall-asguard:
Review Error for dnyu @ 2019-06-04 05:09:40 UTC

PR #42, opened 2019-03-08
updated gems and bundle version issue

nishils:
This update fixes a spec that requires bundler 2 which we do not support yet.

I tried adding bundler 2 support but there are some bugs namely I can't get bundler 2 and bundler 1 to work within the same container yet.

The easy fix is to change the spec for now and fix the rubocop errors so that the project is in a relatively clean state.

Also, performed some housekeeping on updating out of date gems.


lib/salus.rb
```diff
@@ -55,6 +55,7 @@ def scan(
         processor.export_report
       rescue StandardError => e
         raise e if ENV['RUNNING_SALUS_TESTS']
+
```
cbsisl: Any reason a new line was added here?

nishils: These changes were due to rubocop

lib/salus/formatting.rb
```diff
@@ -4,22 +4,22 @@ module Formatting
...
+      top_left: '┌',
+      top_right: '┐',
+      bottom_mid: '┴',
+      bottom_left: '└',
       bottom_right: '┘',
-      left_mid:     '├',
-      right_mid:    '┤',
-      horizontal:   '─',
-      vertical:     '│',
-      cross:        '┼'
```
jborrey: I'm actually a big fan of lining things up vertically. Can we disable the rubocop rule that's forcing this?

jborrey:
LGMT - one comment.

as3richa:
I was curious about the Bundler 2 issue, so I hacked on it a bit and came up with https://github.com/coinbase/salus/pull/45.

Bundler 2 is actually backwards-compatible, so there's no need to install or depend upon both versions in the Salus container. There was a bit of a strange issue in which running `gem install bundler` as part of the Dockerfile successfully installed Bundler 2, but any subsequent calls to `bundle` would be run against Bundler 1. This was fixed by bumping this base image on which we're building.

This PR should go out first, then I can quickly rebase my branch

nishils:
Thanks for figuring that out. I couldn't figure out how to fix that behavior. Anyways, this PR will clean up the gems and rubocop errors.

PR #44, opened 2019-03-11
[chore] remove spurious check on return value from yarn audit scanner

as3richa:
See title. This check never did what was intended; the exit code of yarn audit can be several different values for a successful run (https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/audit.js#L141-L146) none of which are semantically meaningful for our purposes.


jborrey:
@as3richa should we close this or are you still working on it?

as3richa:
This change is both complete and necessary. I just rebased; feel free to merge if/when CI goes green.

(Tangentially, I've left Coinbase in the interim, so this may be my last Salus PR :grin:)

nishils:
Thanks for all your help @as3richa!

I will roll this into the salus 2.4.2 update

PR #45, opened 2019-03-11
[chore] Upgrade to bundler 2

as3richa:


xolaz:
Can you add a description with problem being addressed plus approach?

 Also, how come the ruby containers changing and bundle audit?

nishils:
@as3richa Can we add a test for this? Should be able to copy the existing fixture and change the bundled version in the sample gemfile.lock file.

as3richa:
> Can you add a description with problem being addressed plus approach? Also, how come the ruby containers changing and bundle audit?

This PR upgrades Bundler to version 2 so that we can cleanly run Salus against Ruby projects bundled with Bundler 2. We were pinned to a very old Ruby container that didn't play nice with Bundler 2; this more recent revision works fine. Bundle audit 0.6.0 depends explicitly on Bundler 1, whereas 0.6.1 works cleanly with Bundler 2.

PR #48, opened 2019-04-18
Pin to gosec 1.2.0

nishils:
Right now gosec 1.3.0 is causing us issues because golang is throwing compile errors.

We currently don't have version pinning so this adds it. We will re-evaluate 1.3.0 when we get a chance


heimdall-asguard:
Review Error for xolaz @ 2019-04-18 00:26:02 UTC

PR #30, opened 2019-01-15
[chore] bump Go to 1.11.4

as3richa:

PR #50, opened 2019-04-20
updated dockerfile to support gomodules and newer version of go

nishils:
waiting on https://github.com/securego/gosec/issues/234 so that gosec supports modules

This PR uses go modules and a more modern version of go.


jborrey:
NB: If we merge this, we should close #30.

PR #54, opened 2019-05-23
CircleCI integration documentation

jsulinski:
Here is a usage example for the CircleCI Orb. If you'd like, I can add a more detailed doc to the docs directory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,24 @@ Salus also parses dependency files and reports on what libraries and version are
 
 If you would like to build customer scanners or support more languages that are not currenclty supported, you can use [this method of building custom Salus images](docs/custom_salus.md).
 
+## CircleCI Integration
+
+Salus can be integrated with CircleCI by using a public orb. All Salus configuration options are supported, and defaults are the same as for Salus itself. 
+
+You can do so by adding the following lines to your CircleCI `config.yml`:
+
+```
+version: 2.1
+
+orbs:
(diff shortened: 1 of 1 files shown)


jborrey:
This is slick! +1 from me.

nishils:
@jsulinski This is awesome. Sorry for taking so long to get to it.

Would it make sense to eventually converge the two projects so the versioning between the orb and salus can be the same? Happy to include this for now and re-visit in the future.

jsulinski:
No worries! The other option is to make the 'latest' tag the default. The primary reason I didn’t was because of this suggestion: https://github.com/coinbase/salus/pull/41#issuecomment-487819029. 

I’m happy to go either route and help maintain. If we converge, I can prepare a PR. Do you think it would be better to drop it into a top level directory called 'integrations' (or similar), or a directory beneath docs?

nishils:
I think converging would be a good idea. As this could, let us publish a new salus version with the corresponding orb at the same time.

I agree the orb should not default to latest but maybe providing a non-default option is the way to go. I am not sure if this is possible as I haven't played around with orbs yet.

Yea, it might make sense to throw this in `integrations/orb` or `integrations/circleci` ? Then, we can throw all the other CI configs in there too like Jenkins and Travis whenever that is explicitly supported.

jsulinski:
We can close this merge request in favor of: https://github.com/coinbase/salus/pull/59

PR #55, opened 2019-05-31
Patternmatch config option for including extension

raphdev:
The `pattern_search` (sift) scanner has an option to exclude file extensions from matches, useful for targeting specific files for which you don't want potential false positive matches. This option maps to sift's [`--exclude-ext`](https://sift-tool.org/samples) command line flag. Similarly, sift has an option to only scan specified extensions, but this option is not exposed via the Salus config. This means one has to manually blacklist all undesired extensions.

This PR exposes an `include_extension` salus config option for the pattern_match scanner. In cases where we know exactly which extensions we want to target, this allows us to better target matches.

### DIFF
--- a/docs/scanners/pattern_search.md
+++ b/docs/scanners/pattern_search.md
@@ -2,6 +2,8 @@
 
 This scanner can flag anti-patterns found in a codebase or require that certain strings be present. This might be useful for preventing the use of dangerous methods like `eval()` in Ruby (which might allow for RCE) or `dangerouslySetInnerHTML` in React (which might allow for XSS). By default, all found patterns are added to the info section of the report. If a found pattern is forbidden, this scanner will fail and the `message` will be show to the developer in the report to give additional context on why this was an issue. A `required` pattern must be found in order for the scan to pass.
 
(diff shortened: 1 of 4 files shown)


docs/scanners/pattern_search.md
```diff
@@ -15,8 +15,11 @@ scanner_configs:
         forbidden: true
         exclude_directory:
           - node_modules
-        exclude_extension:
```
nishils: Can we use exclude and include extension simultaneously for different file types or can you only exclude or only include? Can you add a note in the documentation which one takes precedence in case a file extension is both included and excluded?

raphdev: No problem! Will update to more explicitly describe these behaviors. You can use both - while the [documentation](https://sift-tool.org/docs) isn't explicit about precedence, the [specs](https://github.com/coinbase/salus/pull/55/files#diff-57b4177c5dd615aad3f26daaca52e3f9R285) show that exclusion takes precedence over inclusion.

nishils:
Duped in #62 but fixed the rubocop errors there. Will close this PR and merge #62 .

PR #59, opened 2019-06-04
CircleCI orb and documentation

jsulinski:
Orb version is pegged to the Salus version. To use `latest` Docker image tag, see the specify_executor_scan example.


jsulinski:
This was a collaboration with @raphdev (added him as co-author on commit). Previous commit history is here: https://github.com/federacy/salus-orb

PR #63, opened 2019-06-07
2.5.0 bump

nishils:


README.md
```diff
@@ -65,7 +65,7 @@ Example CircleCI `config.yml`:
 version: 2.1
 
 orbs:
-  salus: federacy/salus@2.4.2
+  salus: federacy/salus@2.5.0
```
xolaz: did a new orb get deployed already?

aunyks: [Doesn't look like it, yet.](https://circleci.com/orbs/registry/orb/federacy/salus)

nishils: Not yet. @jsulinski how do we deploy new orbs?

jsulinski: It's bound to a CircleCI account. I'll deploy a new version now!

jsulinski: Added a PR here: https://github.com/coinbase/salus/pull/64

jsulinski: Published.

jsulinski: Happy to add anyone that wants to publish. Alternatively, we could use `latest` as the default tag to remove the need to publish pegged versions.

littledatas: Yeah let's do latest as default!

I would prefer to have one less step to maintain!

jsulinski: Okay, I'll make that change this weekend!

jsulinski: PR for this is here: https://github.com/coinbase/salus/pull/66. Orb has already been pushed, so starting with 2.5.1 it will always pull latest unless you specify a custom version for an executor.

PR #67, opened 2019-06-10
2.5.1 patch

nishils:
This PR fixes the lack of recursive gosec scanning which was made apparent when 2.5.0 made 0 lines of scanned a hard fail.

### DIFF
--- a/integrations/circleci/README.md
+++ b/integrations/circleci/README.md
@@ -4,7 +4,7 @@
 
 | attribute | description | default | options |
 | --------- | ----------- | ------- | ------- |
-| salus_executor | CircleCI executor to use that specifies Salus environment | `coinbase/salus:2.5.0` | See [executor reference](https://circleci.com/docs/2.0/configuration-reference/#executors-requires-version-21)|
+| salus_executor | CircleCI executor to use that specifies Salus environment | `coinbase/salus:2.5.1` | See [executor reference](https://circleci.com/docs/2.0/configuration-reference/#executors-requires-version-21)|
 | active_scanners | Scanners to run | all | Brakeman, PatternSearch, BundleAudit, NPMAudit |
 | enforced_scanners | Scanners that block builds | all | Brakeman, PatternSearch, BundleAudit, NPMAudit |
 | report_uri | Where to send Salus reports | file://../salus-report.json | Any URI |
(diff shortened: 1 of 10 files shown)


lib/salus/scanners/gosec.rb
```diff
@@ -9,7 +9,17 @@ def run
       # Shell Instructions:
       #   - -fmt=json for JSON output
       #   - gosec can scan go modules as of 2.0.0.
-      shell_return = run_shell("gosec -fmt=json #{@repository.path_to_repo}")
+      shell_return = Dir.chdir(@repository.path_to_repo) { run_shell("gosec -fmt=json ./...") }
+
+      # This produces no JSON output so must be checked before parsing stdout
+      if shell_return.stdout.blank? && shell_return.stderr.include?('No packages found')
```
xolaz: can you add a test?

spec/lib/salus/scanners/gosec_spec.rb
```diff
@@ -18,7 +18,7 @@
         expect(
           info[:stderr]
         ).to include(
-          'blank_repository' # debug information
```
xolaz: no packages found is a bit confusing to me, it doesn't necessarily tell me there is not code found, aka blank repo, can we stick with the blank repo description?

lib/salus/scanners/gosec.rb
littledatas: Yeah, the blank project covers this case. It's what found the issue in the first place

integrations/circleci/README.md
```diff
@@ -38,7 +38,7 @@ workflows:
 version: 2.1
 
 orbs:
-  salus: federacy/salus@2.5.0
+  salus: federacy/salus@2.5.1
```
xolaz: I think this will break if we bump the orbs reference if we bump before it's live.

spec/lib/salus/scanners/gosec_spec.rb
littledatas: This comes from the gosec error logs. I think blank_repository also came from the gosec logs.

What do you suggest here?

lib/salus/scanners/gosec.rb
xolaz: one with a go project that requires a recursive scan?

nishils: That makes sense. I have added it

spec/lib/salus/scanners/gosec_spec.rb
nishils: This is also stated in the message. The stderr is just additional output from the scanner itself.

PR #70, opened 2019-08-03
Fix Yarn Audit findings bug and patch-bump version from 2.6.0 -> 2.6.1

geraldnash-cb:
type=nonroutine
risk=low
impact=sev5


lib/salus/scanners/node_audit.rb
```diff
@@ -63,7 +63,7 @@ def run
           # For all advisories,
           prod = raw_advisories_for_id.any? do |raw_advisory|
             # any there there any instances in the prod dependency tree?
-            raw_advisory.fetch(:findings).any? { |finding| !finding.fetch(:dev) }
+            raw_advisory.fetch(:findings).any? { |finding| !finding.fetch(:dev, false) }
```
xolaz: is the fix to not run development dependencies?

geraldnash-cb: No, previously this scanner expected the advisories fetched by YarnAudit to have a `dev` key in each object of their `findings` array. Some advisories don't have the `dev` key and cause the scanner to error on this line.  

The value of the `dev` key is typically `true` or `false`, and this change treats the value as `false` if the key is not provided.

TLDR: It treats the finding as a non-dev finding by default if not already specified. @xolaz

xolaz: Interesting. Can you write a quick test for this?

xolaz: The test can check our advisory output is as expected even without the dev key

xolaz: And what does the dev key mean for NPM/Yarn? That the finding is related to a development package?

geraldnash-cb: @xolaz I'm unable to find documentation related to the shape of these advisory objects and the meaning of their keys. Comments in the code claim that the key lets us know whether advisories are in some production dependency tree, but it's unclear how the scanner would know that and what the difference between the production dependency tree and trees in other staging environments would be, if any.

geraldnash-cb: Drafting a test case right now.

geraldnash-cb: Also, I spoke with @nishils and he said that the difference between production and development advisories is that prod advisories are advisories related `dependencies` and dev advisories are related to `devDependencies`. So, the `dev` key lets us know whether the finding is related to a dev dependency.

geraldnash-cb: So, in essence if we're unsure whether a finding is related to a dev or non-dev dependency (no dev key provided), we assume the worst by treating it as if it's related to a non-dev dependency. @xolaz

nishils: You can also quickly test this @geraldnash-cb by running `yarn audit --json` to see what the output looks like.

nishils:
This is a good bug fix! Can we also add an option to exclude dev dependencies?

geraldnash-cb:
@nishils @xolaz Please let me know once we've published these updates.

PR #76, opened 2019-09-11
CircleCI support for custom_info

jsulinski:
Utilizes custom_info to store CircleCI variables. 

Also bump CircleCI Orb to 3.0.0 to disassociate from Salus version as the Orb is updated far less frequently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ Example CircleCI `config.yml`:
 version: 2.1
 
 orbs:
-  salus: federacy/salus@2.6.2
+  salus: federacy/salus@3.0.0
 
 workflows:
   main:
--- a/integrations/circleci/README.md
+++ b/integrations/circleci/README.md
@@ -14,6 +14,29 @@
 
 Note: active_scanners and enforced_scanners must be yaml formatted for Salus configuration file.
 
+## CircleCI Environment Variables
+
+Stored in custom_info of a Salus scan.
+
+| Key | CircleCI Variable | Description |
+| --- | ----------------- | ----------- |
+| sha1    | CIRCLE_SHA1 | Hash of last commit in build |
+| ci_project_username | CIRCLE_PROJECT_USERNAME | SCM username of project |
+| reponame | CIRCLE_PROJECT_REPONAME | Name of repository |
+| branch | CIRCLE_BRANCH | Name of git branch being built |
+| tag | CIRCLE_TAG | Name of tag |
+| repository_url | CIRCLE_REPOSITORY_URL | URL of the Github or Bitbucket repository |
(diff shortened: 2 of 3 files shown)


nishils:
@jsulinski thanks for the update. If you can provide instructions on how to update the orb, I am more than happy to keep that in my release process. I will push this in 2.6.3

jsulinski:
@nishils There was a discussion in this [PR](https://github.com/coinbase/salus/pull/63) regarding utilizing the `latest` tag to avoid having to push the orb for every release of Salus. The problem is that because I had initially adopted the Salus version number for the orb, it was being bumped in the documentation along with the Salus version bump when there were no changes or releases for the orb. 

This PR resolves the issue by bumping the Orb version to 3.0.0. It also adds support for hashes in custom_info, which will bomb without the corresponding [change](https://github.com/coinbase/salus/pull/76) to Salus itself.

Regarding how to push the orb, instructions are here: https://circleci.com/docs/2.0/creating-orbs/#using-the-cli-to-publish-orbs

However, it might be better for you to create a salus organization, as CircleCI uses the VCS's name and membership for an ACL. There seem to be limited options for renaming an organization, and we worked around this by using a Bitbucket account to nab the federacy namespace. Consequently, for me to give you permissions, you would need to join a Bitbucket repo (and I would need to upgrade from a free account). 

Just let me know if you'd prefer to take this later route, or if you have any questions regarding the former!

nishils:
Sorry for the delayed response, I was a bit heads down this past week.

>@nishils There was a discussion in this PR regarding utilizing the latest tag to avoid having to push the orb for every release of Salus. The problem is that because I had initially adopted the Salus version number for the orb, it was being bumped in the documentation along with the Salus version bump when there were no changes or releases for the orb.

Got it. Thanks for clarifying.

> (and I would need to upgrade from a free account).

And that's how they get you :P 

That's fine no need to pay for that. If you have it pointed to latest, then that should be fine. We don't typically make breaking changes, and when we do, they are usually caused by upstream dependencies making breaking changes.

I appreciate your help maintaining the orb. Getting more features in Salus has been my main focus either better integration with existing scanners or new scanners. Any help I can get to maintain the nice-to-haves is great.

Thanks for your help as always!

jsulinski:
No problem at all, happy to help. Thanks for creating an awesome project and making it open source! Hoping to start working on another integration soon.

If there were breaking changes, it's simple enough for the user to utilize a custom executor to peg to a specific version as well.

jsulinski:
We should merge https://github.com/coinbase/salus/pull/76 along with this as the orb improvements depend on custom_info supporting a hash.

nishils:
Yea, I plan to release 2.6.3 tomorrow officially once I can get a few people to +1 the changes.

PR #94, opened 2020-02-01
Add brakeman config

mrcoinbase:
This is to add a brakeman config, let me know if the configuration builder and the documentation is good. I want to add some integration testing around this.


mrcoinbase:
Tests are failing due to https://github.com/rubysec/bundler-audit/pull/236

PR #95, opened 2020-02-04
Mr add brakeman config

mrcoinbase:
Adds a brakeman config and also a way of adding configs to other scanners quickly in the future.


docs/scanners/brakeman.md
```diff
@@ -5,10 +5,46 @@ The [Brakeman Scanner](http://brakemanscanner.org/) is a static analysis tool th
...
 Brakeman's configuration is complex and it parses its own `brakeman.ignore` file. Salus will let Brakeman reuse this file if it's present. To create one, you can use `brakeman -I` which lets you run an interactive scan.
+For more information on Brakeman configs, see [Brakeman Options](https://brakemanscanner.org/docs/options/)
 
 ## Configuration
 ```yaml
   scanner_configs:
     Brakeman:
-    - path: path/to/rails/app             # default -  looks for project_root/app
+        c: "path/to/config" # Config file for brakeman path, anything in the command line config here will override the brakeman config file. By default it will look for a config in: ./config/brakeman.yml, ~/.brakeman/config.yml, and /etc/brakeman/config.yml
+        A: true # When true, Brakeman runs all checks, off by default
```
nishils: For these single letter options, is it possible to use a more verbose name that is a bit more intuitive? 

If its something we make our own translation for like "`config` is the `c` brakeman config option which defines the config file for brakeman patch..."

mrcoinbase:
@nishils I definitely want to check that the documentation is clear on how to use this, since it would be nice to plug and play for future scanners we add

lib/salus/formatting.rb
```diff
@@ -4,22 +4,22 @@ module Formatting
     INDENT_STRING = (' ' * INDENT_SIZE).freeze
 
     TABLE_BORDER_CHARS = {
-      top_mid:      '┬',
-      top_left:     '┌',
-      top_right:    '┐',
-      bottom_mid:   '┴',
-      bottom_left:  '└',
+      top_mid: '┬',
```
nishils: I'm not sure why rubocop flagged these. These should have been marked as an exception. I think for this file we prefer the way it is stylistically.

lib/salus/report.rb
```diff
@@ -14,7 +14,7 @@ class ExportReportError < StandardError; end
     CONTENT_TYPE_FOR_FORMAT = {
       'json' => 'application/json',
       'yaml' => 'text/x-yaml',
-      'txt'  => 'text/plain'
+      'txt' => 'text/plain'
```
nishils: same here as the above rubocop comment

docs/scanners/brakeman.md
mrcoinbase: That would make sense, but then we'd fragment from the way the scanner works, so we'd vary from brakeman's documentation.

lib/salus/formatting.rb
mrcoinbase: Yeah, I ran rubocop -a

docs/scanners/brakeman.md
mrcoinbase: Actually, let's run with that. I'll make that change

lib/salus/scanners/base.rb
```diff
@@ -178,10 +179,201 @@ def validate_file_option(keyword)
...
+      return '' unless validate_bool_option(keyword, value)
+
+      if value.to_s.downcase == "true"
+        "#{prefix}#{keyword}#{suffix}"
+      else
+        ''
+      end
+    end
+
+    def create_bool_option(keyword:, value:, prefix:, between:, suffix:)
```
nishils: nit: Perhaps a better name for `between` like `keyval_separator`

mrcoinbase: How about just separator?

nishils: nit: `suffix` like `arg_separator`

nit: `prefix` like `option_separator`

Probably fine though

lib/salus/scanners/node_audit.rb
```diff
@@ -12,10 +12,10 @@ class NodeAudit < Base
 
     # Some common, easily-abbreviated terms to shorten advisory titles
     TITLE_SUBSTITUTIONS = {
-      'Regular Expression'    => 'Regex',
-      'Denial of Service'     => 'DoS',
+      'Regular Expression' => 'Regex',
```
nishils: I'm not sure why rubocop is messing up all over the place here

spec/lib/salus/scanners/brakeman_spec.rb
```diff
@@ -35,6 +35,26 @@
         expect(logs).to include('Dangerous Eval')
       end
 
+      it 'should respect the config for ignoring files' do
```
nishils: there seems like there should be more tests for all the config options.

Writing the integration tests are a pita but help us make sure the underlying scanners don't break underneath us when doing an upgrade.

lib/salus/formatting.rb
nishils: hmm i typically do that but wonder why the ignores didn't trigger for you

lib/salus/scanners/base.rb
nishils: Well, there's a lot of separators here though. Naming is hard enough and this only has to be remembered once anyways that it isn't a huge deal

mrcoinbase:
@nishils put in extra tests, ability to add more descriptive names, undid rubocop, and renamed 'between' to 'separator'

docs/scanners/base.md
```diff
@@ -34,3 +34,72 @@ scanner_configs:
...
+
+```-flag -string=foo --list=foo,bar,baz -bool=true -file=./foobar.js -file_list=foo.js,bar.js -multiple first -multiple second -d```
+
+Note each type for the arguments, the supported types are:
+```ruby
+  :flag
+  :string # Numbers qualify as strings
+  :list # For a list of strings
+  :bool
+  :file
```
nishils: can there be more  examples? Do they need to be prepended by `./`? What is the path file relative to?

docs/scanners/base.md
```diff
@@ -34,3 +34,72 @@ scanner_configs:
...
+For example, let's pretend you intend to send in the configurations in the following format:
+
+```-flag -string=foo --list=foo,bar,baz -bool=true -file=./foobar.js -file_list=foo.js,bar.js -multiple first -multiple second -d```
+
+Note each type for the arguments, the supported types are:
+```ruby
+  :flag
+  :string # Numbers qualify as strings
+  :list # For a list of strings
+  :bool
```
nishils: case sensitive or not? Provide examples of expected values please!

nishils: there is something called :booleans referred to below. Does this need to be documented?

lib/salus/scanners/base.rb
```diff
@@ -124,64 +125,263 @@ def report_dependency(file, hsh = {})
...
+                         suffix: type_value[:suffix] || suffix,
+                         separator: type_value[:separator] || separator,
+                         keyword: type_value[:keyword] || keyword_string,
+                         value: config_value,
+                         type: type_value[:type],
+                         regex: type_value[:regex] || default_regex,
+                         join_by: type_value[:join_by] || join_by
+                       )
+                     end
+                   when Regexp # Assume it is a string type if just regex is supplied
```
nishils: Is this true? Can I provide a regex against file matching?

*.js for example?

nishils:
left a few nits for docs but  this looks good. looks like you  need to fix the rebase conflicts

lib/salus/scanners/base.rb
mrcoinbase: Yeah, you could, but it would need to be /.*\.js/

The only difference between a string and a file type in these is whether or not the code checks if the file already exists.

docs/scanners/base.md
mrcoinbase: :bool = :boolean, I just made it an alias

mrcoinbase: also, these are not case sensitive. all get normalized to downcase

docs/scanners/base.md
mrcoinbase: ? Those are all the cases ?

PR #104, opened 2020-03-26
ruby 2.7.0 upgrade

ghbren:


.circleci/config.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       - run: docker run -t -v $(pwd):/home/repo coinbase/salus
   rubocop:
     docker:
-      - image: ruby:2.4.6@sha256:3a31984805c5ad3b54baeb93d2c01c46845f681b712394b02d2e860cb5d5946b
+      - image: ruby:2.7.0@sha256:b516edb5cfe1091b6dabdfe97180f3bcc6c01c3439500d5ab720b46a3e75d5da
```
xolaz, line 18: where are you grabbing the sha from? 🤔

xolaz:
can you also update the .ruby_version file? https://github.com/coinbase/salus/blob/master/.ruby-version#L1

.circleci/config.yml
ghbren: this was the output of "docker pull ruby:latest"

PR #106, opened 2020-04-03
add semgrep scanner

DrewDennison:
@nishils, @xolaz , myself, and some of the other [sgrep](https://sgrep.dev/) maintainers discussed augmenting Salus' regex-based `PatternSearch` scanner with one that uses syntactic grep. The goal of this PR is to prototype an sgrep-based scanner to improve the consistency and quality of user-defined checks.

[`sgrep`](https://sgrep.dev) (syntactic grep) is an open-source tool for finding patterns in code. It's useful for preventing the use of known anti-patterns in a codebase or enforcing the correct use of secure-by-default frameworks (e.g. always use a project's sanitization method on user-provided data). It's grep-esque patterns are lifted into AST matchers.

```python
flask.response.set_cookie(...)
```

will match on

```python
response.set_cookie("hello", "world")

response.set_cookie("hello", # This is a comment
                    "world")
```

You can see more example patterns and experiment yourself with [sgrep.live](https://sgrep.live) 

We hope that with sgrep it’s easier for Salus users to enforce their custom code standards.

### Overview
This PR contains:

1. Add a v0 minimal integration of sgrep
2. Add basic documentation
3. Add rspec tests

### Testing Plan
Manual testing along with rspec integration tests were done. I validated with the docker-based testing instructions.

### Open Questions
 
 - [ ] Are there Salus-specific customizations or missing functionality that you’d like to see added?
 - [ ] What more sample or test patterns would be helpful or valued?


ghbren:
Does sgrep support searching in Go comments?  For example, can I search for all comments that include "hello_world", like "# 1234 hello_world 5678"?

DrewDennison:
@ghbren good question. We don't because we lift the code into a AST and drop the comments. I suspect PatternSearch is better for searching comments but would love to learn more about what you're trying to accomplish or if it's a suppression syntax you're aiming for

nishils:
This is super cool! Definitely in support of getting this added. I'll defer to @xolaz and @ghbren on if anything is missing. I've transitioned out of being a core maintainer and more of an active contributor.

Dockerfile
```diff
@@ -63,6 +63,19 @@ RUN go get github.com/svent/sift@$SIFT_VERSION \
 RUN go get github.com/securego/gosec/cmd/gosec@915e9ee \
   && mv /root/go/bin/gosec /usr/bin/
 
+### sgrep tool install https://sgrep.dev
+ENV SGREP_VERSION 0.4.9b5
```
DrewDennison, line 67: I'm going to cut a stable release but wanted to get feedback on the overall PR first

xolaz:
thanks @DrewDennison, @ghbren's reviewing your PR. We'll circle back this week.

ghbren:
@DrewDennison I have reviewed the PR and have some questions.

1) Is there a way to specify a whitelist?  For example, some scanners allow you specify a list of directories, files, or file extensions to exclude.  If Sgrep supports it, could you add that to the documentation?

2) For each pattern, the user is required to specify the language, which makes sense if there are multiple languages in the repo.  But most repos have just one language.  Can Sgrep auto detect the language in such repos?

3) The error message for unsupported languages is a bit misleading.  For example, if I specify an unknown language, then the error is  "stderr:  in manual, pattern in rule - can't be parsed for language".  Perhaps the error message should instead be that the language is not supported?

4) If a forbidden pattern have multiple matches, then the only error message I get is the the user-specified message in config.  Can the scanner also print out the relevant meta data, such as file paths, line numbers, and relevant code?

Thank you.

DrewDennison:
hey @ghbren thanks for the thoughtful questions. I'm happy to get them addressed this week.

1. WIP and will be done shortly. I'm planning to follow the convention in `pattern_search.rb`
```
#   - exclude_directories: Array of directories (GLOB) in the repo to exclude from the search.
#   - exclude_extensions: Array of file extensions to exclude from the search.
#   - include_extensions: Array of file extensions to scan exclusively.
#   The above can also be provided per-match, and will override the global values. Exclusions
#   take precedence over inclusions if they conflict.
```

2. Good feature request. Not at the moment but it's on the roadmap. Would it work to have a single global language option for the config with overrides for now? 

3. Errors are getting cleaned up this week and are looking _much_ nicer in my local dev builds of sgrep
4. Yes I have a TODO to add the line to the output. Do you want just one line or would you like a config for num lines of context a la grep.

Action Items

- [ ] @DrewDennison to add include/exclude and update the tests to include tests for that i.e. use the same tests from PatternSearch
- [ ] @DrewDennison to add support for language setting at the scanner level?
- [ ] @DrewDennison to update the version with better errors and line context (should be ready in a few days)

Anything I missed? Does the overall structure feel about right? Any other comments/blockers?

ghbren:
Hi @DrewDennison ,

1) and 3) sound great.

2) perhaps sgrep could auto detect the language of the repo by looking for files such as go.mod or Gemfile? It would also be useful to allow the user to override the language for each pattern in the repo config.

4) I think it would be useful to show the file path, line numbers, and the piece of code that matches the pattern. But perhaps sometimes the piece of code may be too large, so a config option for the first n lines would useful.

The overall structure feels right, and no other comments/blockers for now.  I also asked about a search in comments feature, but that's probably not needed, so we can ignore that for now.

Thank you very much

DrewDennison:
@ghbren I updated the PR with what I believe support for all the requested features. Happy to discuss more here or in slack.

docs/scanners/semgrep.md
```diff
@@ -0,0 +1,22 @@
...
+
+semgrep is fast and powerful; it's grep-esque patterns are lifted into AST matchers. Compared to regexes these patterns aren’t affected by whitespaces, comments, newlines, the order of keyword arguments, variable renaming, and other language nuances.
+
+If a found pattern is forbidden this scanner will fail and the `message` will be show to the developer in the report. A `required` pattern must be found in order for the scan to pass
+
+## Configuration
+
+```yaml
+scanner_configs:
+  config:
```
ghbren, line 13: config appears to be a typo. It should be Semgrep instead.

docs/scanners/semgrep.md
```diff
@@ -0,0 +1,22 @@
+# [semgrep](https://semgrep.dev)
+
+[`semgrep`](https://semgrep.dev) (syntactic grep) is an open-source tool for finding patterns in code. It's useful for preventing the use of known anti-patterns in a codebase or enforcing the correct use of secure-by-default frameworks (e.g. always use a project's sanitization method on user-provided data).
+
+semgrep is fast and powerful; it's grep-esque patterns are lifted into AST matchers. Compared to regexes these patterns aren’t affected by whitespaces, comments, newlines, the order of keyword arguments, variable renaming, and other language nuances.
+
+If a found pattern is forbidden this scanner will fail and the `message` will be show to the developer in the report. A `required` pattern must be found in order for the scan to pass
+
```
ghbren, line 8: Could you add the languages supported?  Also, add some info on config options like --exclude-dir, --exclude, --include, like in https://github.com/coinbase/salus/blob/master/docs/scanners/pattern_search.md?

ghbren:
@DrewDennison Were you going to add support for adding info like file paths and line numbers to matches?

DrewDennison:
@ghbren that should work already. See https://github.com/coinbase/salus/pull/106/files#diff-a3cee78d09fa30467a918e93fbbf29d4R37 for an example output

ghbren:
@DrewDennison Thanks for pointing that out.  I see how info[:hits] shows the line number in spec code. But when I run semgrep from salus directly, it doesn't print out line numbers.

DrewDennison:
That's strange. Can you send me the config you're using so I can debug locally?

ghbren:
@DrewDennison Here's an example. If I print out the shell_return, I do see the pattern line number.  But when I docker run salus directly, the line number is not shown.

```
def foo(x):
    x.unsanitize(123)

# salus.yaml
scanner_configs:
  Semgrep:
    matches:
      - pattern: $X.unsanitize(...)
        message: unsanitzed stuff
        language: python        
        forbidden: true
```

ghbren:
@DrewDennison  
Also, if I add 
```
        include_extension:
          - "*.js"
          - "*.py"
```
to the above salus.yaml, then sgrep finds no matches.  Is this a bug?

DrewDennison:
@ghbren fixed the path and line context in the direct run and verified that by running `docker salus` with 

```
Semgrep:
    matches:
      - pattern: $X.unsanitize(...)
        message: unsanitzed stuff
        language: python        
        forbidden: true
        include_extension:
          - py
          - js
        exclude_directory:
          - example
```

I also reproduced, fixed, and made the format of `{include,exclude}_extension` match `PatternSearch`.

ghbren:
@DrewDennison Thank you for checking in the bug fixes. But now I am confused about the relationship between `language` and `included_extension`.  If I set language to python but include js extension only.  Will sgrep still search for the pattern in js code?  I tried it out but sgrep didn't find the forbidden pattern in my js file.

DrewDennison:
@ghbren file filtering is still a bit messier than I would like. At the moment, there are 2 stages. if language is `python`, it will only search `.py` files and the same goes for other languages. Specifically for this PR (to make it more closely match PatternSearch), I threw in glob filtering for `include_extention`, `exclude_extention`, and `exclude_directory` that serves as a pre-filter. Happy to remove them but since you asked for them happy to have it in and I'm unifying that code in the next few releases. Happy to hop on a quick slack call and discuss live if that helps clarify things and work out any remaining FRs

ghbren:
@DrewDennison Hi Drew, can you tell me a little about the expected behavior of `include/exclude extension` with `language`?   Does `include_extension` have higher precedence over `language`?  What's the expected behavior of `include_extension: js` used with `language: python`?

DrewDennison:
sure, include/exclude filter things at the file walking level. The language flag defines the AST pattern syntax for parsing etc. So if you do `include_extension: js` you will only get Javascript files and then trying to run with `language: python` means it will only search python ASTs which is will results in no matches. So your specific query above would have no matches. Personally, I've only ever used the language flag and that seems to work well but the other flags are designed to enable filtering out large folders of irrelevant code which might help with performance at coinbase scale

ghbren:
@DrewDennison Thanks for the explanation.  Is that a language that matches multiple file extensions in semgrep?

DrewDennison:
@ghbren there is not. There is support for an array of languages but I didn't want to complicate the PR with that. In practice, I have never written a rule for more than 1 language because there is just too much code specific details but the same principles apply so it's more of a learn once, write everywhere.

ghbren:
@DrewDennison Since every language matches exactly one file extension, I suppose it makes sense to remove the `--include/exclude-extension` flags?   I was thinking these extension flags might cause some confusions to our users.  

After that, I think the PR would be ready for merging.

DrewDennison:
@ghbren cool makes sense

docs/scanners/semgrep.md
```diff
@@ -0,0 +1,26 @@
...
+```yaml
+scanner_configs:
+  Semgrep:
+    matches:
+      - pattern: $X == $X
+        message: Useless equlity check
+        language: python
+        exclude_directory:
+          - tests
```
ghbren, line 21: @DrewDennison Is the user required to specify `required` or `forbidden` for each pattern?  This pattern above is missing that keyword.

DrewDennison: @ghbren I don't think so. I based my PR on https://github.com/coinbase/salus/blob/master/lib/salus/scanners/pattern_search.rb#L54-L55 but of course it's easy to make it so. Let me know what you prefer.

ghbren: @DrewDennison It looks like that pattern above doesn't do anything, because the pattern is not `required` nor `forbidden`, correct?  If so, could you just add either `forbidden: true` or `required: true` to that example?

ghbren:
@DrewDennison I was doing some final testing and found another issue.
It appears that `exclude_directory` does not work with subdirectories.
For example, I created this `salus.yaml`, and `subdir/subdir2` contains test.py that has the forbidden pattern
```
scanner_configs:
  Semgrep:
    matches:
      - pattern: $X.foo(...)
        message: should not call foo(...)
        language: python        
        forbidden: true
        exclude_directory:
           - subdir
```
Semgrep passes in this case as expected, because `subdir/subdir2/test.py` has been excluded in the config.

Now, if I change the `exclude_directory` from `subdir` to `subdir/subdir2`, then semgrep fails and reports forbidden pattern found in `subdir/subdir2/test.py`.  But shouldn't semgrep pass in this case?

docs/scanners/semgrep.md
DrewDennison: For sure. My understanding of that was it would log but not break anything but `forbidden` means it breaks the build and `required` breaks the build if it doesn't fire but I trust you know Salus much better than I do.

DrewDennison:
> @DrewDennison I was doing some final testing and found another issue.
> It appears that `exclude_directory` does not work with subdirectories.
> For example, I created this `salus.yaml`, and `subdir/subdir2` contains test.py that has the forbidden pattern
> 
> ```
> scanner_configs:
>   Semgrep:
>     matches:
>       - pattern: $X.foo(...)
>         message: should not call foo(...)
>         language: python        
>         forbidden: true
>         exclude_directory:
>            - subdir
> ```
> 
> Semgrep passes in this case as expected, because `subdir/subdir2/test.py` has been excluded in the config.
> 
> Now, if I change the `exclude_directory` from `subdir` to `subdir/subdir2`, then semgrep fails and reports forbidden pattern found in `subdir/subdir2/test.py`. But shouldn't semgrep pass in this case?

I've been using `exclude_directory` as a way to filter out vendored code deeply nested in our monorepo. It's been pretty useful for filtering out all `test` and `node_modules` folders but I can see how this is confusing; I updated the readme clarifying that `subdir/subdir2` isn't supported. It would be helpful if there is a spec for what `exclude_directory` should support. Would you prefer `exclude_directory` being .gitignore syntax, regex, prefix match, or full recursive glob such as `**/node_modules/**`?If you have a tool that does this particularly well in mind, happy to copy the API from that.

ghbren:
@DrewDennison I think the current syntax is fine.

@xolaz Do you think `exclude_directory` needs to support nested directories? If not, then I think the semgrep scanner is ready to be merged.

xolaz:
@ghbren I don't see it as a blocker. Now, let's note in the config readme so behavior is clear to users of Salus and semgrep. 

In general, being able to exclude nested directories is helpful for a lot of mono repos especially.

DrewDennison:
Yeah I'll add to my todo list to support more expressive excludes. Thanks for the feedback

PR #111, opened 2020-04-09
added Acronym for Salus

ghbren:
Added acronym for Salus in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
 
 ## 🔍 Overview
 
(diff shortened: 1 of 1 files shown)


jborrey:
Since when was this a thing?

ghbren:
> Since when was this a thing?

Should it be something else?

jborrey:
I just don't recall there ever being an acronym.
When I named Salus, it was just the Roman Goddess.

If you want some actual trivia, it was going to be called Samus (after Samus Aran) but then I didn't want to think about the copyright infringement etc etc. Salus was a good alternative.

The acronym might help people understand the goal quickly though, so 👍 .

xolaz:
@jborrey we jokingly made one up back in Jan and then realized it fits in really nicely with what Salus does.

nishils:
Yea, this is definitely a backronym

PR #112, opened 2020-04-09
Bren/bugsnag

ghbren:
Added bugsnag integration documentation and notify on report warn/error/send

### DIFF
--- a/lib/salus.rb
+++ b/lib/salus.rb
@@ -64,6 +64,7 @@ module Salus
         raise e if ENV['RUNNING_SALUS_TESTS']
 
         puts "Could not send Salus report: (#{e.class}: #{e.message})"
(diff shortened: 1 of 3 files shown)


lib/salus.rb
```diff
@@ -64,6 +64,7 @@ def scan(
         raise e if ENV['RUNNING_SALUS_TESTS']
 
         puts "Could not send Salus report: (#{e.class}: #{e.message})"
+        Bugsnag.notify(e)
```
xolaz, line 67: What if someone doesn't have bugsnag integrated? Will this fail for them?

xolaz:
can you update this branch with master?

lib/salus.rb
ghbren: No, it will not fail.

PR #118, opened 2020-04-21
only call Bugsnag.notify if user provided Bugsnag API key

ghbren:


lib/salus.rb
```diff
@@ -7,10 +7,10 @@
     config.endpoint = ENV.fetch('BUGSNAG_ENDPOINT', 'notify.bugsnag.com')
     config.api_key = ENV['BUGSNAG_API_KEY']
   end
-end
 
-# Hook at_exit to send off the fatal exception if it occurred
-at_exit { Bugsnag.notify($ERROR_INFO) if $ERROR_INFO }
+  # Hook at_exit to send off the fatal exception if it occurred
+  at_exit { Bugsnag.notify($ERROR_INFO) if $ERROR_INFO }
```
xolaz, line 12: any reason we use `at_exit` for this 🤔?  I didn't notice this before.

lib/salus.rb
```diff
@@ -64,7 +64,7 @@ def scan(
         raise e if ENV['RUNNING_SALUS_TESTS']
 
         puts "Could not send Salus report: (#{e.class}: #{e.message})"
-        Bugsnag.notify(e)
+        Bugsnag.notify(e) if ENV['BUGSNAG_API_KEY']
```
xolaz, line 67: hmm, definitely fixes the warning. Can we aim for a centralized solution? Maybe create a wrapper method too replace all the `Bugsnag.notify(e) if ENV['BUGSNAG_API_KEY']` checks

```
def log_error(e)
  if ENV['BUGSNAG_API_KEY']
    Bugsnag.notify(e)
  else
    # log to a generic logger
  end
end
```

ghbren: added wrapper Salus.bugsnag_notify to wrap the code in the if branch in your example. 
The else branch code is not added because the original code logs messages differently depending on the situation (ex. @report.warn vs @report.error)

lib/salus.rb
ghbren: $ERROR_INFO represents unhandled exceptions.  This only place to ensure all such exceptions are reported is the program exit point.
This call has been in salus since the first commit of salus.rb.

heimdall-asguard:
This Pull Request was merged without enough reviews
Please go [here](https://github.com/coinbase/salus/issues/122) to review and +1

PR #121, opened 2020-04-27
fix wrong number of arguments passed to validate_file_option by makin…

ghbren:
…g 2nd argument optional

### DIFF
--- a/lib/salus/scanners/base.rb
+++ b/lib/salus/scanners/base.rb
@@ -160,19 +160,7 @@ module Salus::Scanners
       false
     end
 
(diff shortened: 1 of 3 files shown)


lib/salus/scanners/base.rb
```diff
@@ -160,7 +160,19 @@ def validate_list_option(keyword, value, regex)
       false
     end
 
-    def validate_file_option(keyword, value)
+    def validate_file_option(keyword, value = -1)
```
xolaz, line 163: can you write a test for this?

PR #143, opened 2020-06-15
allow for special chars in regexes for PatternSearch 

DrewDennison:
and not have them escaped by the shell


ghbren:
@DrewDennison Thank you very much for submitting the PR.  I see that you have a spec test showing your changes fixed the problem.  But somehow when I run salus with the docker command, it shows PatternSearch passed even when forbidden patterns are found.

This incorrect behavior is likely unrelated to your changes.  It is probably caused by a bug with how salus processes salus.yaml.  I will investigate that first and then come back to this PR.

DrewDennison:
sounds good. Yeah I can't see why that would change based on my PR but happy to fix if it is for some reason

PR #147, opened 2020-06-23
Gosec update

ghbren:
* Upgrade gosec to 2.3.0
* Remove `go get` from gosec scanner because it is not needed if using go modules with gosec 2.3.0


Dockerfile
```diff
@@ -64,9 +64,7 @@ RUN curl -fsSL "$GOLANG_DOWNLOAD_URL" -o golang.tar.gz \
 RUN go get github.com/svent/sift@$SIFT_VERSION \
   && mv /root/go/bin/sift /usr/bin/
 
-# GOSEC SHA Commit 915e9ee (915e9eeba8982f1e8ec82186900dff2d729e6dd4)
-RUN go get github.com/securego/gosec/cmd/gosec@915e9ee \
-  && mv /root/go/bin/gosec /usr/bin/
+RUN curl -sfL https://raw.githubusercontent.com/securego/gosec/master/install.sh | sh -s -- -b /usr/bin v$GOSEC_VERSION
```
xolaz, line 67: can we point to a gosec sha?

PR #150, opened 2020-06-24
disallow some keywords to be specified in both salus.yaml and semgrep…

ghbren:
Disallow `pattern`,  `language`, and `message` for the same rule to be defined in both salus.yaml and semgrep config.
Values in semgrep config currently overrides salus.yaml, and may cause confusions to the user.


lib/salus/scanners/semgrep.rb
```diff
@@ -39,6 +39,19 @@ def run
           # semgrep has the following behavior:
           #   ouputs a json object with a 'results' and 'errors' key
 
+          if !match['config'].nil?
+            override_keys = %w[pattern language message]
```
cbtoni, line 43: Is it `language` or `languages`? It also looks like there is both `pattern` and `patterns` allowed in the semgrep config 🤔

lib/salus/scanners/semgrep.rb
```diff
@@ -39,6 +39,19 @@ def run
...
+            override_keys.any? do |k|
```
cbtoni, line 44: I think you mean `override_keys.each` here?

ghbren: yes, thank you

lib/salus/scanners/semgrep.rb
ghbren: * salus.yaml only supports `language`, semgrep yaml only supports `languages`. 
  - semgrep yaml requires `languages` and will override `language` from salus.yaml
* salus.yaml only supports `pattern`, semgrep yaml supports `pattern` and `patterns`
  - semgrep yaml requires either `pattern` or `patterns`, and will override `pattern` from salus.yaml

lib/salus/scanners/semgrep.rb
ghbren: @cbtoni I just addressed the comments and updated error message.  Could you re-review? Thanks

PR #157, opened 2020-06-30
Add information about missing configs to report

cbtoni:
- Updated the processor to distinguish between configs that were valid and used, and configs that were not. 
- Updated some comments to make it more clear about the difference between the implicit config name (`salus.yaml`) and the default or base config (`salus-default.yaml`).
- I added some output to the normal report, not sure if its the best way to do it so lmk if you have other suggestions. It now adds an extra section that looks like this:
```
==== Salus Configuration Used:

  file:///salus.yaml
  file:///test_config.yml
```

### DIFF
--- a/lib/salus/processor.rb
+++ b/lib/salus/processor.rb
@@ -7,6 +7,8 @@ module Salus
 
     attr_reader :config, :report
 
+    # If configuration sources are not provided, we'll automatically scan the
+    # repo root for a configuration file with this default name
     DEFAULT_CONFIG_SOURCE = "file:///salus.yaml".freeze
 
     def initialize(configuration_sources = [], repo_path: DEFAULT_REPO_PATH)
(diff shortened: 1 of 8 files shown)


lib/salus/processor.rb
```diff
@@ -7,27 +7,42 @@ class InvalidConfigSourceError < StandardError; end
...
       # Add default file path to the configs if empty.
       configuration_sources << DEFAULT_CONFIG_SOURCE if configuration_sources.empty?
+      valid_sources = []
+      source_data = []
 
       # Import each config file in order.
-      files = configuration_sources.map { |source| fetch_config_file(source, repo_path) }
+      configuration_sources.each do |source|
+        body = fetch_config_file(source, repo_path)
+        if !body.nil?
```
ghbren, line 25: Does it make sense to output an error/warning if body is nil?

cbtoni: I thought about that, but since we look for `salus.yaml` by default, if you run this on a repo without a config it will always output an error?

PR #160, opened 2020-07-14
bump semgrep to 0.14.0 and make salus compatible with 0.14.0

ghbren:
* Bump semgrep to 0.14.0
* Update salus semgrep to  work with the `--exclude` config option, which supercedes the old `--exclude-dir`. 
* Add `rule_id` to patterns found output.  This can be useful with the new whitelisting feature.
* Add whitelisting info in semgrep doc.
* Update salus semgrep to be compatible with new semgrep output format changes. 
  *  ```
        # old semgrep parser error format

        "message": "syntax error ....",
        "details": {
          "path": "junk.js",
          "start": {
             "line": 1,
             "col": 1
          },
          "end": {
             ....
          },
          "message": "...",
          "line": ...
        }
      } 

  * ```
        # new semgrep parser error format

          "type": "SourceParseError",
          "short_msg": "parse error",
          "long_msg": "Could not parse junk.js as javascript",
          "level": "warn",
          "spans": [
            {
              "start": {
                "line": 1,
                "col": 1
              },
              "end": {
                ...
              },
              "source_hash": "...",
              "file": "doh.js",
            }

### DIFF
--- a/lib/salus/scanners/semgrep.rb
+++ b/lib/salus/scanners/semgrep.rb
@@ -207,7 +207,7 @@ module Salus::Scanners
     def message_from_hit(hit, match)
       has_external_config = !match['config'].nil?
       msg = if has_external_config
-              hit['extra']['message']
+              hit['extra']['message'] + "\n\trule_id: " + hit['check_id']
             else
               match['message']
             end
--- a/lib/salus/scanners/semgrep.rb
+++ b/lib/salus/scanners/semgrep.rb
@@ -5,7 +5,7 @@ require "json"
 # semgrep is a grep like tool for doing AST pattern matching.
 # See https://semgrep.dev for more info and documentation.
 # Config file can provide:
-#   - exclude_directories: Array of directories in the repo to exclude from the search.
+#   - exclude: Skip any file or directory that matches this pattern;
 #   - language: A global language flag if you don't want to set it per pattern.
 #   The above can also be provided per-pattern, and will override the global values.
 #   - matches: Array[Hash]
@@ -23,7 +23,7 @@ require "json"
 module Salus::Scanners
   class Semgrep < Base
     def run
-      global_exclude_directory_flags = flag_list('--exclude-dir', @config['exclude_directory'])
+      global_exclude_flags = flag_list('--exclude', @config['exclude'])
 
       # For each pattern, keep a running history of failures, errors, warnings, and hits
       # These will be reported on at the end.
@@ -58,18 +58,19 @@ module Salus::Scanners
           match["strict"] ||= false
           match["message"] ||= ""
 
-          pattern_exclude_directory_flags = flag_list(
-            '--exclude-dir', match['exclude_directory']
+          pattern_exclude_flags = flag_list(
+            '--exclude', match['exclude']
           )
 
           command, user_message = build_command_and_message(
             match,
             @config['strict'] || match["strict"],
             base_path,
-            pattern_exclude_directory_flags || global_exclude_directory_flags
+            pattern_exclude_flags || global_exclude_flags
           )
           # run semgrep
           shell_return = run_shell(command)
+
           # check to make sure it's successful
           if shell_return.success?
             # parse the output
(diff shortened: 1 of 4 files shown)


lib/salus/scanners/semgrep.rb
```diff
@@ -229,25 +227,34 @@ def hit_to_string(hit, base_path)
...
-        line: line
+        level: level,
+        spans: spans
       }
     end
 
     def error_to_string(err)
       err_obj = error_to_object(err)
-      "#{err_obj[:message]}\n\t#{err_obj[:path]}:#{err_obj[:start].fetch('line', '')}"
+      spans = err_obj[:spans].map do |s|
```
xolaz, line 254: what is the var `spans` getting used for in this method?

lib/salus/scanners/semgrep.rb
```diff
@@ -209,7 +207,7 @@ def build_command_and_message(match, strict, base_path, exclude_directories_flag
     def message_from_hit(hit, match)
       has_external_config = !match['config'].nil?
       msg = if has_external_config
-              hit['extra']['message']
+              hit['extra']['message'] + "\n\trule_id: " + hit['check_id']
```
xolaz, line 210: I'm trying to understand this change. Can you walk me through it? 🤔

ghbren: In the new semgrep version, there is a whitelisting feature that allows you to whitelist rule ids.
But in the previous salus version, if a forbidden pattern is found, it doesn't tell you what the rule id is, like

```
  Forbidden patterns in config "semgrep_rules.yaml" was found 
  	test.js:      password = "hardcoded_secret"```
```

So I changed the output to include the rule_id, like
```
  Forbidden patterns in config "semgrep_rules.yaml" was found 
        rule_id: hardcoded_secret
  	test.js:      password = "secret"
```

This way, the user can whitelist the rule id like


```
password = "secret" // nosem: hardcoded_secret
```

cbtoni: Does the `check_id` need to be in the message text? It seems like it would be more useful as a separate field. Also, it's not clear why you are checking for `has_external_config` to determine if the message should be changed?

ghbren: @cbtoni, did you mean make the output look like below?
{code}
  Forbidden patterns in config "semgrep_rules.yaml" was found 
  	test.js:10:      password = "secret"
  Forbidden patterns in config "semgrep_rules.yaml" was found 
  	test.js:20:     if (0 == 0) ...

  Forbidden pattern rule_ids: hardcoded_secret, useless comparison
{code}

`has_external_config` is used when you reference semgrep config files in salus.yaml.  Too add a semgrep rule, you can either reference a semgrep config file in salus.yaml, or add the rule directly. But if you add the rule directly, it only works with simple rules and the `rule_id` option is not supported.

cbtoni: I think I'm a little confused about how the messages are put together in the output from reading the specs, and I thought there would be one entry for each hit, with a message and a rule_id. If that's not the case then you can ignore me.

ghbren: @cbtoni There is one entry for each hit, with a message and a rule_id, like below. How did you want `rule_id` to be a separate field in the output below?

```
==== Semgrep: FAILED in 2.87s

 ~~ Scanner Logs:


  Forbidden patterns in config "semgrep_rules.yaml" was found - Hardcoded password 1
  	rule_id: hardcoded_password_1
  	pass.js:1:password =

  Forbidden patterns in config "semgrep_rules.yaml" was found - Plesae use safeLoad if possible
  	rule_id: unsafe_yaml_load_3
  	c/t.js:4:      return <pre className="ex">{yaml.dump(yaml
  .load(definition.get("example")), defaultYamlOptions)}</pre>;

```

cbtoni: @ghbren I think it's fine like that, thanks for sharing an example with me

PR #164, opened 2020-07-21
Rust spike

joshuaostrom:
Wanted to familiarize myself w/the code base and saw there was an open request for Rust support.  This is just a quick spike - I included a scanner.  Should be straight forward to add dependency tracking if desired.

Also touched the brakeman_spec.rb to relax a few hard coded paths that were causing failing specs locally ( I assume fine when ran w/in docker, should be g2g now in either environment).


ghbren:
@joshuaostrom Thank you for submitting the PR.  

You listed 3 cargo audit behaviors: 1) vulnerability not found, 2) vulnerability found, and 3) error running audit.  Does `shell_return.success?` return true if and only if 1)? 

Could you address the following?

* add a `should_run?` method.  This method should return true if there is a Cargo.lock.
* add support for config options. Please see how `config_options` is defined and used in gosec, brakeman, and bandit.
* finish cargo_audit.md
* fix failed circle ci tests

Feel free to let me know if you have any questions.

joshuaostrom:
Thanks for the feedback!  I missed the comment notification so I’m a bit belated here in seeing the comments.

Traveling at the moment, but will aim to address the feedback here.

joshuaostrom:
Thanks @ghbren ! I've pushed up some updates.  Going to address the various questions:

> You listed 3 cargo audit behaviors: 1) vulnerability not found, 2) vulnerability found, and 3) error running audit. Does shell_return.success? return true if and only if 1)?

Correct, only vulnerability not found gives back a status code of zero / success? == true

> Could you address the following?
> 
> add a should_run? method. This method should return true if there is a Cargo.lock.
This is still present / untouched from the original PR:

```
    def should_run?
      @repository.cargo_lock_present?
    end
```
If there are any concerns there just let me know - happy to take a look

> add support for config options. Please see how config_options is defined and used in gosec, brakeman, and bandit.

This has now been added. 

> finish cargo_audit.md

This has now been updated.

> fix failed circle ci tests

These have been updated and are now passing

As noted on the original PR notes I have the rust scanner, I can add the rust dependency reporting too - I assume we'd like that

joshuaostrom:
I've added dependency reporting.  PR is ready for review.

README.md
```diff
@@ -38,6 +38,8 @@ docker run --rm -t -v $(pwd):/home/repo coinbase/salus
 - [npm audit](docs/scanners/npm_audit.md) - Execution of [`npm audit`](https://docs.npmjs.com/getting-started/running-a-security-audit) 6.13.7 which looks for CVEs in node module dependencies.
 - [yarn audit](docs/scanners/yarn_audit.md) - Execution of [`yarn audit`](https://yarnpkg.com/lang/en/docs/cli/audit/) 1.22.0 which looks for CVEs in node module dependencies.
 - [PatternSearch](docs/scanners/pattern_search.md) - Execution of [`sift`](https://sift-tool.org/docs) 0.9.0, looks for certain strings in a project that might be dangerous or could require that certain strings be present.
+- [Cargo Audit](docs/scanners/cargo_audit.md) - Execution of [Cargo Audit](https://github.com/RustSec/cargo-audit) Audit Cargo.lock files for crates with security vulnerabilities reported to the RustSec Advisory Database
```
ghbren, line 41: could you add the cargo audit version, like the other scanners?

Dockerfile
```diff
@@ -25,6 +25,11 @@ RUN apt-get update && apt-get upgrade -y --no-install-recommends && apt-get inst
     wget \
   && rm -rf /var/lib/apt/lists/*
 
+# Rust
+RUN curl https://sh.rustup.rs -sSf | bash -s -- -y
+ENV PATH="/root/.cargo/bin:${PATH}"
+RUN cargo install cargo-audit
```
ghbren, line 31: Add SHA check?

lib/salus/scanners/cargo_audit.rb
```diff
@@ -0,0 +1,96 @@
...
+        #
+        # - Error running audit:
+        #   - status: 1
+        #   - stderr: String with details on the error prenting the run (not JSON)
+        #   - stdout: ""
+
+        return report_success if shell_return.success?
+
+        if shell_return.stderr.empty?
+          report_failure
```
ghbren, line 36: Can you move `report_failure` to right before `if shell_return.stderr.empty?`
Like gosec https://github.com/coinbase/salus/blob/master/lib/salus/scanners/gosec.rb#L38

Otherwise, the `else` branch will not be flagged as a failure.

lib/salus/scanners/cargo_audit.rb
```diff
@@ -0,0 +1,96 @@
...
+      #     --stale                   allow stale database
+      #     --target-arch TARGET-ARCH filter vulnerabilities by CPU (default: no filter)
+      #     --target-os TARGET-OS     filter vulnerabilities by OS (default: no filter)
+      #     -u, --url URL             URL for advisory database git repo
+      #     -q, --quiet               Avoid printing unnecessary information
+      #     --json                    Output report in JSON format
+      #     --no-local-crates         Vulnerability querying does not consider local crates
+
+      opts = ["--json"]  # return vulnerabilities in an easily digestible manner (JSON)
+      opts << "-c never" # to prevent color chars in stderr upon failure
```
ghbren, line 74: Is there a reason we don't want to show color chars?

README.md
joshuaostrom: Sounds good, I've locked the version of cargo audit and added that version info (0.12.0) to the Readme

Dockerfile
joshuaostrom: Rust provides PGP signatures, I've added these to our dockerfile

lib/salus/scanners/cargo_audit.rb
joshuaostrom: Ah good catch - we call report_failure for both vulnerabilities and failed scans.  Added tests cases for the expanded behavior.

lib/salus/scanners/cargo_audit.rb
joshuaostrom: Just for cleaner logging:

`error: Couldn't load Cargo.lock
`
vs

`\e[0m\e[0m\e[1m\e[31merror:\e[0m Couldn't load Cargo.lock
`

That said, if we *want* the markup in the error log just let me know it's easy enough to drop that -c never option.

ghbren (approved):

PR #166, opened 2020-08-07
add file location to forbidden pattern output

ghbren:
Add file locations to output


ghbren:
will reopen this PR from a branch

PR #167, opened 2020-08-07
add file locations to forbidden pattern output

ghbren:
Add file locations to forbidden pattern output

Old output
```Forbidden pattern "dangerous_call" was found ```

New output
```
Forbidden pattern "dangerous_call" was found 
myfile.py:123:     dangerous_call
```


ghbren:
made a mistake in PR, will re-open

PR #174, opened 2020-08-12
support exclude_filepaths

ghbren:
Add support for exclude_filepaths option.


spec/fixtures/pattern_search/test_paths/salus.yaml
```diff
@@ -0,0 +1,10 @@
+scanner_configs:
+  PatternSearch:
+    matches:
+      - regex: hello
+        forbidden: true
+        exclude_filepaths:
```
xolaz, line 6: can you exclude filepaths and directories at the same time? If so, can you make a test case for it?

ghbren: Yes, just added test case

spec/fixtures/pattern_search/test_paths2/salus.yaml
```diff
@@ -0,0 +1,12 @@
...
+    matches:
+      - regex: hello
+        forbidden: true
+        exclude_extension:
+          - html
+        exclude_filepaths:
+          - file.rb
+          - file.py
+        exclude_directory:
+          - subdir
```
xolaz, line 12: what if the directory is nested, what's the format for that? can you update the readme with that detail at least?

ghbren: updated the doc for `exclude_directory`

PR #176, opened 2020-08-18
add support for build_url in salus config

ghbren:


ghbren:
will make another PR that all allows customized build info in salus config

PR #181, opened 2020-09-03
parse yarn audit text output into json

ghbren:


lib/salus/scanners/yarn_audit.rb
```diff
@@ -7,12 +7,78 @@
 
 module Salus::Scanners
   class YarnAudit < NodeAudit
-    AUDIT_COMMAND = 'yarn audit --json'.freeze
```
cbtoni, line 10: Can we leave a comment here, maybe add a link to https://github.com/yarnpkg/yarn/issues/7404 so we remember why we changed this?

lib/salus/scanners/yarn_audit.rb
```diff
@@ -7,14 +7,84 @@
...
+        shell_return = run_shell(command)
+
+        excpts = @config.fetch('exceptions', []).map { |e| e["advisory_id"].to_i }
+        report_info(:ignored_cves, excpts)
+        return report_success if shell_return.success?
+
+        stdout_lines = shell_return.stdout.split("\n")
+        table_start_pos = stdout_lines.index(TT)
+        table_end_pos = stdout_lines.rindex(TB)
+        table_lines = stdout_lines[table_start_pos..table_end_pos]
```
cbtoni, line 32: I think you should move this after we check if `table_start_pos.nil?`. Is there a test for the case where there is a `stderr` return but no vulns/tables in the output?

lib/salus/scanners/yarn_audit.rb
```diff
@@ -7,14 +7,84 @@
...
+        if lines[i] == TT
+          vuln = {}
+        elsif lines[i] =~ /^\│ [A-Za-z]/
+          # line has attr name and val, like
+          # | Path          | eslint > file-entry-cache > flat-cache > write > mkdirp >    |
+          line_split = lines[i].split("│")
+          key = line_split[1].strip
+          val = line_split[2].strip
+          vuln[key] = val
+        elsif lines[i].start_with?(empty_cell)
```
cbtoni, line 73: Instead of looking for lines with `empty_cell`, perhaps you could just check if the `key` is empty after you do `.strip` on it?

lib/salus/scanners/yarn_audit.rb
```diff
@@ -35,19 +105,41 @@ def scan_deps
       command << 'optionalDependencies ' unless dep_types.include?('optionalDependencies')
     end
 
-    def scan_for_cves
-      # Yarn gives us a new-line separated list of JSON blobs.
-      # But the last JSON blob is a summary that we can discard.
-      # We must also pluck out only the standard advisory hashes.
-      command = "#{AUDIT_COMMAND} #{scan_deps}"
-      command_output = run_shell(command)
+    # severity and vuln title in the yarn output looks like
```
cbtoni, line 108: Nice comment!

lib/salus/scanners/yarn_audit.rb
cbtoni: Can we also add a test to make sure we are capturing multi-line values correctly?

cbtoni (changes requested):
A few comments, I'm also going to run this on some test repos too

lib/salus/scanners/yarn_audit.rb
```diff
@@ -7,14 +7,84 @@
 
 module Salus::Scanners
   class YarnAudit < NodeAudit
-    AUDIT_COMMAND = 'yarn audit --json'.freeze
+    # the command was previously 'yarn audit --json', which had memory allocation issues
+    # see https://github.com/yarnpkg/yarn/issues/7404
+    AUDIT_COMMAND = 'yarn audit --no-color'.freeze
+    TT = "┌───────────────┬──────────────────────────────────────────────────────────────┐".freeze
+    TB = "└───────────────┴──────────────────────────────────────────────────────────────┘".freeze
```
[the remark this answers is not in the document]
ghbren: The tables always have the same width. But I just updated the code to check the first characters instead.

lib/salus/scanners/yarn_audit.rb
ghbren: FIx the `table_start_pos.nil?`

ghbren: Yes, there is already a test case for the case where there is a stderr return but no vulns/tables in the output.
It is `should fail with error if there are errors` in yarn_audit_spec.rb.

lib/salus/scanners/yarn_audit.rb
ghbren: Good suggestion. I got rid of the `empty_cell`.

ghbren: The spec test on `failure-2` dir already covers multi-line values.
If you run `yarn audit` directly on the dir, it shows a multi-line value.

cbtoni (approved):
Seems to work for me locally!

PR #182, opened 2020-09-08
remove dependency paths from yarn audit output

ghbren:
Remove `Path` from yarn audit output, and sort results by ID.


lib/salus/scanners/yarn_audit.rb
```diff
@@ -66,19 +66,22 @@ def parse_output(lines)
...
         elsif lines[i].start_with?("└─") && lines[i].end_with?("─┘")
-          vulns.push vuln
+          vulns.add(vuln)
         end
         i += 1
       end
 
-      vulns.each { |vln| normalize_vuln(vln) }
+      vulns = vulns.to_a
+      vulns.each { |vln| normalize_vuln(vln) }.sort { |a, b| a['ID'] <=> b['ID'] }
```
xolaz, line 84: can you add a test for this?

ghbren: Good suggestion!  Test added.

xolaz (approved):

PR #187, opened 2020-09-18
Removed race condition and added missing require

joshuaostrom:
An omitted require in CargoAudit (2.10.11) caused a failure in production. This was masked by a race condition in the loading order. The require as been added and the loading is no longer random.

### DIFF
--- a/lib/salus/scanners.rb
+++ b/lib/salus/scanners.rb
@@ -1,6 +1,7 @@
 module Salus::Scanners; end
 
-Dir.entries(File.expand_path('scanners', __dir__)).each do |filename|
+# Sort to avoid race coniditions in what order scanners are loaded
+Dir.entries(File.expand_path('scanners', __dir__)).sort.each do |filename|
   next if ['.', '..'].include?(filename) # don't include FS pointers
   next unless /\.rb\z/.match?(filename)  # only include ruby files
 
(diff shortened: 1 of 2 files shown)


ghbren (dismissed):

lib/salus/scanners.rb
```diff
@@ -1,6 +1,7 @@
 module Salus::Scanners; end
 
-Dir.entries(File.expand_path('scanners', __dir__)).each do |filename|
+# Sort to avoid race coniditions in what order scanners are loaded.
+Dir.entries(File.expand_path('scanners', __dir__)).sort.each do |filename|
```
xolaz, line 4: how did you determine sorting as the issue here?

joshuaostrom: This block of code dictates the order in which the various scanner classes are loaded.  Ruby doesn't guarantee the order in which Dir.entries will return files.

Manually placing cargo_audit.rb as the first entry reproduced the production error. 

By sorting, we are removing the varying order in which classes may load - and the race conditions that can occur if there is any order dependence.  This ensures that as we run executions across our various pipelines (dev, testing, production) we will remove discrepancies in the load process.  The sort itself doesn't matter rather it's just a way to ensure consistent loading between various environments.

joshua-ostrom:
Closing this as Circle was not running tests.  Opened a new PR that is running tests.

PR #197, opened 2020-09-25
Plugin spike for user-friendly config overrides

joshua-ostrom:
PR to discuss opening up a plugin architecture.  In practice, if we want to move forward we'd want to do some light refactoring to Salus to support various filters, classes, hooks via plugins.

Some low hanging fruit might be telemetry (statsd plugin, etc), report plublishers (kafka, etc), config processors.  Expanding the plugin process to also simply allowing gems names in the salus config etc.

Opening up Salus with a plugin architecture might afford a clean approach to expanding the feature set and a good catalyst for public extensions to Salus.   This PR is provides a quick spike to show how we could plugin custom config logic.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -90,6 +90,10 @@ scanner_configs:
   BundleAudit:
     ignore:
       - CVE-XXXX-YYYY # irrelevant CVE which does not have a patch yet
+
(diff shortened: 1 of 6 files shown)


cbtoni:
I'm going to close this PR for now as we further explore the designs for plugins in Salus.

PR #208, opened 2020-10-06
Persist scanner versions

ghbren:
Updated salus to persist version for each scanner as part of Salus report output.

Now salus stdout shows versions
```
...
==== Brakeman 4.9.1: PASSED in 3.36s

==== BundleAudit 0.6.1: PASSED in 0.8s 

==== PatternSearch 0.9.0: PASSED in 0.0s
...
```

Same for json output
```
    ...
    "PatternSearch": {
      "scanner_name": "PatternSearch",
      "version": "0.9.0",
      "passed": true,
     ...
    "Semgrep": {
      "scanner_name": "Semgrep",
      "version": "0.14.0",
      "passed": true,
    ...
```


heimdall-asguard:
If Heimdall is stuck, please click the following link: https://heimdall-public.cbhq.net/v1/pull_request/stuck?pr_number=208&repository=coinbase/salus

lib/salus/scan_report.rb
```diff
@@ -130,7 +133,7 @@ def render_banner(use_colors:)
       status = passed? ? 'PASSED' : 'FAILED'
       status = colorize(status, (passed? ? :green : :red)) if use_colors
 
-      banner = "==== #{@scanner_name}: #{status}"
+      banner = "==== #{@scanner_name} #{@version}: #{status}"
```
cbtoni, line 136:
```suggestion
      banner = "==== #{@scanner_name} v#{@version}: #{status}"
```

It might be nice for the stdout text to include a `v` in front of the version number for clarity and to also to match the other stdout lines like `==== Salus Scan v2.10.13 for ...`. 

This is a nitpick though 😄 ⛏️

lib/salus/scanners/base.rb
```diff
@@ -14,6 +14,18 @@ class InvalidScannerInvocationError < StandardError; end
...
+      'Bandit' => '1.6.2',
+      'Brakeman' => '4.9.1',
+      'BundleAudit' => '0.6.1',
+      'CargoAudit' => '0.12.0',
+      'Gosec' => '2.4.0',
+      'NPMAudit' => '6.14.8',
+      'PatternSearch' => '0.9.0',
+      'Semgrep' => '0.14.0',
+      'YarnAudit' => '1.22.0'
+    }.freeze
```
joshua-ostrom, line 27: Thoughts on pulling these in from the shell (or gem env) to report the version that will run?

```
`brakeman --version`
=> "brakeman 4.9.1"
cargo audit --version`
=> "cargo-audit 0.12.0"
```
The ones that have gems we can pull dynamically as well:
```
Gem.loaded_specs["brakeman"].version.to_s
=> "4.9.1"
```


etc.  Right now we define the version in the docker container or gem file.  If we bump the docker image but forget to bump the version here we might end up with some mis-versioned reports.

On the base scanner we might want to define a 

```
def version
  raise 'Subclass should supply version
end
```

Then in the subclass
```
class BrakemanScanner < Base
def version
  Gem.loaded_specs["brakeman"].version.to_s
end
end

```
(or self.version if you want to pull these from the class )

This would help encapsulate all the subclasses logic in the subclass (and minimize the number of places we have scanner specific code).

joshua-ostrom (changes requested):
See comments on where/how we define the scanner versions

lib/salus/scanners/base.rb
ghbren: @joshua-ostrom That's a very good idea.  I'm totally fine with implementing what you suggested.

A related issue is that we currently hard-code the scanner versions in the README. Perhaps we could automate that too.

@cbtoni , what do you think?

cbtoni: Seems fine. I'm not sure its worth trying to automate that in the readme though.

ghbren: @joshua-ostrom , @cbtoni .  
I'm going to implement Josh's idea of adding a `version` method to each scanner.  But instead of adding 
```
def version
  raise 'Subclass should supply version
end
``` 
to the base scanner, I think it makes sense to make the base scanner return empty string.  This is because some scanners have no versions, like ReportNotEmpty.  It's also hard to get versions from some scanners, like DataTheorem, because we do not have direct access to the version.  So for these scanners, it seems good to use empty string as their versions from the base scanner.

Please let me know what you think.

joshua-ostrom: Yeah if we don't want to enforce that scanners supply a version then defaulting in base is fine to me

lib/salus/scan_report.rb
ghbren: I added the `v`.  Here's how the stdout looks like now
```
==== BundleAudit v0.6.1: PASSED in 2.99s

==== Gosec v2.4.0: PASSED in 1.99s

==== PatternSearch v0.9.0: PASSED in 0.0s

==== RepoNotEmpty: PASSED in 0.0s

==== Semgrep v0.14.0: PASSED in 0.0s

==== YarnAudit v1.22.0: PASSED in 1.08s

==== ReportNodeModules: PASSED in 0.07s

==== ReportRubyGems: PASSED in 0.01s
```

ghbren:
Updated code to get scanner versions dynamically and added tests for getting scanner versions.
If a scanner has no version (like Report* or RepoNotEmpty), then the version is empty string.

spec/support.rb
```diff
@@ -1,3 +1,7 @@
 def remove_file(file_path)
   File.delete(file_path) if File.exist?(file_path)
 end
+
+def version_valid?(version)
```
joshua-ostrom, line 5: We can leverage RSpec custom matchers here to clean up the tests a little.  Helper functions in specs are often a fair signal that you might have an opportunity for a matcher.
This
`expect(version_valid?(scanner.version)).to be_truthy
`
could be
`expect(scanner.version).to be_a_valid_version
`
Given
```
RSpec::Matchers.define :be_a_valid_version do
  match do |actual|
     /^\d+\.\d+(.\d+)*$/ =~ actual
  end
end
```

lib/salus/scanners/gosec.rb
```diff
@@ -54,6 +54,12 @@ def run
       end
     end
 
+    def version
+      shell_return = run_shell('gosec --version')
+      # stdout looks like "Version: 2.4.0\nGit tag: v2.4.0\nBuild date: 2020-07-24T07:54:54Z"
+      shell_return.stdout.split('Version:')[1].split("\n")[0].strip
```
joshua-ostrom, line 60: Food for thought, we *could* access these in a safer fashion:

`shell_return.stdout&.split('Version:')&.dig(1)&.split("\n")&.dig(0)&.strip
`
And then we'd eliminate rescuing from StandError when we send to bugsnag.

spec/support.rb
ghbren: very good suggestion. Modified accordingly.

lib/salus/scanners/gosec.rb
ghbren: modified accordingly

ghbren:
@joshua-ostrom I modified the code based on your comments. Could you review again? Thanks.

joshua-ostrom (approved):
Looks good Brianna.

`!/^\d+\.\d+(.\d+)*$/.match(version).nil?`
could probably be
`version.match?( /^\d+\.\d+(.\d+)*$/)`
for extra clarity but that's minor enough I think you're g2g!

heimdall-asguard:
Review Error for joshua-ostrom @ 2020-10-15 22:19:17 UTC

joshua-ostrom (approved):

PR #214, opened 2020-10-09
ignore running times when comparing reports

ghbren:
In salus spec, there are places where two salus reports are compared.
This sometimes causes spec to fail because the scanner running times are different.

Updated spec so that the running times are ignored.


heimdall-asguard:
If Heimdall is stuck, please click the following link: https://heimdall-public.cbhq.net/v1/pull_request/stuck?pr_number=214&repository=coinbase/salus

joshua-ostrom (commented):
See comments

joshua-ostrom:
I will add, that it would be really nice to have an option to turn off some of the report metadata.  (As we will have the same issue down stream when we do the diff).  That would be another way to tackle this problem... instead of hiding the values in the specs to compare, provide an option to omit them.

Just food for thought.

joshua-ostrom (approved):

PR #234, opened 2020-11-13
send scanner / build info to bugsnag

ghbren:
send scanner / build info to bugsnag


joshua-ostrom:
The changes look good - might be nice to include at least one spec as it doesn't look like we have any checks to ensue we are calling bugsnag_notify

joshua-ostrom (approved):
We lack some spec coverage on bugsnag we can follow w/some coverage there.  PR looks good

ghbren:
bugsnag spec added in #240

Issue #58, opened 2019-06-04
Adding a new scanner checklist


nishils:
Provide a checklist of what every scanner should support in terms of integrating with Salus.

This can include things like 

- please write tests
- please add documentation in the salus docs folder
- please document config in the salus docs folder.

This will make sure that scanners are consistent and make it easier for folks to contribute.

jborrey:
+1

kiwiz:
Hi, looking to contribute and it'd be great if this checklist was completed for referenc.

nishils:
I'm actually about to start working on a custom scanner so I can create a first draft here in the next day or two.

nishils:
@kiwiz - Opened a PR for this. If you'd like to take a look and provide feedback on anything confusing or unclear, it would be greatly appreciated.

PR #255, opened 2020-12-24
resolve issue #58 - adding notes on how to add custom scanner

nishils:
adding notes on how to add custom scanner


kiwiz:
Looks pretty good! Some comments:

- Looks like a documentation requirement is missing from the checklist - is this intentional?
- It'd be helpful to have some recommendations about output should be returned to Salus (of particular interest in the SARIF issue)
- Wrt to scanner maintenance, how are upgrades managed?

nishils:
Good points! I'll get these clarified.

We haven't been two prescriptive on output recommendations though we do have a lot of interest in SARIF so we'll have to re-think this. In general, lots of tools output JSON or provide a JSON output option and we generally just dumped that output.

joshua-ostrom (approved):

PR #241, opened 2020-11-22
Optimize the Dockerfile a bit

mattlorimor:
This PR optimizes some of the layers and their commands. It should shave off roughly 280 MB from the final image size.

PR #257, opened 2020-12-29
Optimize Dockerfile

kiwiz:
- Move around sections && normalize headers
- Normalize env vars and checksum verification
- Combine multiple commands into a single `RUN` command
- Delete all temporary files


kiwiz:
Thoughts on moving the `brakeman` dependency into its own section (as opposed to being embedded in the `Gemfile` as it's a scanner?

kiwiz:
Opps, just realized that #241 exists and has a few distinct changes.

nishils:
Thanks for contributing to this and helping to clean up the Dockerfile! Much appreciated.

We could probably combine some of the small edits in #241 into this PR.

nishils:
> Thoughts on moving the brakeman dependency into its own section (as opposed to being embedded in the Gemfile as it's a scanner?

I like this a lot as it keeps the scanners in a separate place from the core dependencies. However, I'm not sure how easy this would be to do. Would you create another Gemfile like `Gemfile.scanners` and then both of those get installed?  I think the other way to do this is to group the dependencies within the Gem file. What are your throughts?

kiwiz:
Thanks for looking. I just incorporated those changes and forced-pushed up. I'm not too familiar with Ruby, but that approach seems reasonable to me. I can take a stab at that now.

nishils:
So, the nice thing about [bundler](https://bundler.io) is that it pulls in and locks the entire dependency tree. I recommend keeping bundler/gemfiles around.

According to https://bundler.io/v2.2/man/bundle-install.1.html, you can specify the gemfile using the `--gemfile`command. I've never tried it but it looks promising.

nishils:
Actually, I thought about it for a few minutes. The big drawback is that bundler resolves the entire dependency tree in the `Gemfile.lock` file so multiple Gemfiles are going to be weird if there's ever a dependency in core Salus versus a scanner since conflicting dependencies won't be resolved by bundler. The correct way to to this is using groups within the Gemfile (see: https://bundler.io/v2.2/man/gemfile.5.html#GROUPS)

kiwiz:
Alright, moved up Ruby + Yarn deps and made sure all tests passed locally.

nishils:
@kiwiz can we rebase this on master so we can catch the dependency changes? I can +1 and merge then

kiwiz:
@nishils #265 actually contains this changeset so I think we can actually just close this PR out. Thanks for the quick response!

PR #259, opened 2020-12-30
Normalize version checks and remove extraneous spaces

kiwiz:
Due to how multiline strings work in Ruby, the following example actually includes a large number of whitespace:
```ruby
var = "one two\
   four"
# "one two   four"
```
This changeset fixes those (within `base.rb` at least).

I also normalized some version checks within each scanner, added an example of stdout where appropriate & fixed some typos.


ghbren (approved):

ghbren (approved):

heimdall-asguard:
Review Error for ghbren @ 2021-01-29 06:50:19 UTC
User cannot review their own commit

joshua-ostrom (approved):

PR #260, opened 2020-12-30
Integrate Gitleaks scanner

kiwiz:
Work in progress implementation of Gitleaks - pushing up early for feedback.

- Extracts the `sarifLog.runs[]` array and populates it into a `ScanReport` instance
- `ScanReport` now contains some basic code to render SARIF `results` for text output
    - For `json`, it should just emit the raw object
- This is currently on top of #257 - I'd love if that + #241 got merged


ghbren:
Hi, thanks for the PR.  Could you fix the failed circle ci tests?

kiwiz:
Yup, working on adding some tests atm.

kiwiz:
@ghbren Ready for review, but shouldn't be merged until after #269 (if you want to merge this changeset first, I can update the PRs).

I've removed all the SARIF bits as that's still under discussion.

kiwiz:
Rebased against master.

ghbren:
Thanks.  Will probably review next week.  This PR is huge with 51 changed files, so it might take a while to review.

kiwiz:
Rebased and removed some emails.

ghbren:
@kiwiz  Thank you for your PR and apologies for the late reply.  We will do our best to review the PR by the end of the month, but cannot guarantee we will do so.

ghbren (changes requested):
can you add a sha check when installing gitleaks?

lib/salus/scanners/gitleaks.rb
```diff
@@ -0,0 +1,109 @@
...
+        shell_return = run_shell(
+          'gitleaks '\
+          '--path . '\
+          '--format sarif '\
+          '--report /dev/stdout '\
+          '--leaks-exit-code 0 ' +
+          copts
+        )
+
+        report_stderr(shell_return.stderr)
```
ghbren, line 22: `report_stderr` should only be called if there is actually an error.

ghbren (changes requested):

kiwiz:
There have been a few changes to `gitleaks` since I last looked at it. I'm going to take a closer look on Monday & I'll give you a ping once the PR's ready for review.

lib/salus/scanners/gitleaks.rb
```diff
@@ -0,0 +1,109 @@
...
+          depth: :int
+        }
+      )
+    end
+
+    # Report SARIF output from the scanner.
+    def report_runs(runs)
+      # If the results object had findings (workaround for https://github.com/zricethezav/gitleaks/pull/530)
+      if runs.any? { |run| !run['results'].nil? && !run['results'].empty? }
+        @report.fail
```
ghbren, line 89: scanners should not directly call `@report.fail` or `@report.pass`.  
You should use `report_success` for pass and `report_failure` for fail.

ghbren (changes requested):

heimdall-asguard:
Review Error for ghbren @ 2021-04-18 03:43:52 UTC
User failed mfa authentication, public email is not set on your github profile. see [go/mfa-help](http://go/mfa-help)

ghbren:
There are a bunch of binary files in this PR.  What are these? Should they be removed?

docs/scanners/gitleaks.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Configuration
+
+Gitleaks currently always runs with the `--recursive` and `--format json` options turned on.
+In addition, you can configure the additional options below.
+For more information on Gitleaks configuration, see the docs [here](https://github.com/zricethezav/gitleaks#usage-and-options)
+
+```yaml
+  scanner_configs:
+    Bandit:
```
ghbren, line 13: I'm confused... Should this be GitLeaks, not Bandit?

docs/scanners/gitleaks.md
```diff
@@ -0,0 +1,43 @@
...
+      commits:
+      - "aaaaaaaaaaa"
+      - "aaaaaaaaaaa"
+      commits-file: "./commits.txt"
+      commit-from: "aaaaaaaaaaa"
+      commit-to: "aaaaaaaaaaa"
+      commit-since: "2020-01-01T00:00:00-0000"
+      commit-until: "2020-01-01T00:00:00-0000"
+      depth: 1
+```
```
ghbren, line 32: Do your test cases cover each one of the config options here?
What about some combinations of the options?

lib/salus/scanners/gitleaks.rb
kiwiz: Gitleaks reports runtime stats to stderr. Ex:
```
INFO[0001] scan time: 1 second 870 milliseconds 717 microseconds
WARN[0001] leaks found: 14
```
Is there a better place to report this information?

docs/scanners/gitleaks.md
kiwiz: All config options other than `threads` are tested. A number of the `commit*` options are tested together.

kiwiz:
@ghbren Tests should be all fixed now.

> There are a bunch of binary files in this PR. What are these? Should they be removed?
I've embedded a git repo for testing purposes (since `gitleaks` scans individual commits, as well). Open to better ideas here - would it be preferable to `tgz` everything up and extract when running tests?

ghbren:
@kiwiz Is the whole git repo required for testing gitleaks?  Would it be possible to just pick a few relevant files?

kiwiz:
Afaik, the full repo is required to test the commit-scanning functionality. If we remove those tests, then it would be sufficient to just have some files with secrets.

ghbren:
> Afaik, the full repo is required to test the commit-scanning functionality. If we remove those tests, then it would be sufficient to just have some files with secrets.

@kiwiz Does gitleaks scan binary files?  Exactly what are these binary files?

kiwiz:
Yup - gitleaks scans commit objects inside a Git repo. If you want to verify the contents of the commit files, you can rename the `dotGit` directory to `.git` and run normal git commands from there.

spec/lib/salus/processor_spec.rb
```diff
@@ -1,17 +1,19 @@
 require_relative '../../spec_helper.rb'
 
 RSpec::Matchers.define :match_report_json do |expected|
-  def remove_key(json_string, key = 'running_time')
+  def remove_keys(json_string)
```
ghbren, line 4: Could you revert this change?  Removing the extra parameter may break some of our private tests.

kiwiz:
* Rebased against master (kept Golang at 1.15.11)
* Reverted change to `remove_key`

Dockerfile
```diff
@@ -93,6 +93,16 @@ ENV SIFT_VERSION v0.9.0
 RUN go get github.com/svent/sift@${SIFT_VERSION}
 
 
+### gitleaks
+ENV GITLEAKS_VERSION v7.4.1
```
ghbren, line 97: Could you update to the most recent version?

ghbren:
Could you merge with master?  That will fix the weird circle ci test failure.

ghbren:
There are 10+ binary files in the repo.  Are all of them needed to test the commit scanning feature?

docs/scanners/gitleaks.md
```diff
@@ -0,0 +1,43 @@
+# [Gitleaks](https://github.com/zricethezav/gitleaks)
+
+The [Gitleaks Scanner](https://github.com/zricethezav/gitleaks) tool for finding secrets within a directory or Git repo.
+
+## Configuration
+
+Gitleaks currently always runs with the `--recursive` and `--format json` options turned on.
```
ghbren, line 7: Did you mean "always runs" or "by default runs"?
If you specify `--format sarif`, then GitLeaks will produce a sarif only without the json, correct?

docs/scanners/gitleaks.md
```diff
@@ -0,0 +1,43 @@
...
+In addition, you can configure the additional options below.
+For more information on Gitleaks configuration, see the docs [here](https://github.com/zricethezav/gitleaks#usage-and-options)
```
ghbren, line 9: Your scanner only supports the Gitleaks config options below, correct?
If so, can you clarify that and say not all config options in the link above are supported?

docs/scanners/gitleaks.md
```diff
@@ -0,0 +1,43 @@
...
+
+```yaml
+  scanner_configs:
+    Gitleaks:
```
ghbren, line 13: Can you write a brief comment next to each of these config options. For example,
```yaml
config-path: "./gitleaks.conf"   # Path to local gitleaks config
...
```

docs/scanners/gitleaks.md
ghbren: Also, can you specify whether each option is required or optional?

lib/salus/scanners/base.rb
```diff
@@ -178,6 +178,16 @@ def validate_bool_option(keyword, value)
       false
     end
 
+    def validate_int_option(keyword, value)
+      return true if value&.to_s&.match?(/\d+/)
```
ghbren, line 182: Wouldn't this match something like `00001`, which is not exactly a valid int?

lib/salus/scanners/base.rb
```diff
@@ -178,6 +178,16 @@ def validate_bool_option(keyword, value)
...
+
+      report_warn(:scanner_misconfiguration, "Expecting #{keyword} to be a numeric "\
```
ghbren, line 184: Can you update `numeric` to `int`?  because `numeric` includes decimals like `1.234`, but we only want ints.
In fact, maybe `non-negative int`?

lib/salus/scanners/gitleaks.rb
ghbren: What if 0 leaks found? Will Gitleaks still report something like below to stderr?
```
INFO[0001] scan time: 1 second 870 milliseconds 717 microseconds
WARN[0001] leaks found: 0
```

ghbren: We don't really need to report the runtime info.

lib/salus/scanners/gitleaks.rb
```diff
@@ -0,0 +1,109 @@
...
+        )
+
+        report_stderr(shell_return.stderr)
+
+        # A non-zero exit code is definitely an error
+        if !shell_return.success?
+          report_stdout(shell_return.stdout)
+          report_error("Call to gitleaks failed", {
+                         status: shell_return.status,
+                         stderr: shell_return.stderr.split("\n").last
```
ghbren, line 29: Why only report the last in stderr?

lib/salus/scanners/gitleaks.rb
```diff
@@ -0,0 +1,109 @@
...
+          '--report /dev/stdout '\
+          '--leaks-exit-code 0 ' +
+          copts
+        )
+
+        report_stderr(shell_return.stderr)
+
+        # A non-zero exit code is definitely an error
+        if !shell_return.success?
+          report_stdout(shell_return.stdout)
```
ghbren, line 26: can you change `report_stdout` to `report_stderr`?

ghbren: Should you pass the stuff from both `shell_return.stdout` and `shell_return.stderr` to `report_stderr`?

lib/salus/scanners/gitleaks.rb
```diff
@@ -0,0 +1,109 @@
...
+        end
+
+        # Otherwise handle the SARIF output
+        # As of 2.7.1, no report is generated if no secrets were found
+        sarif = if shell_return.stdout.empty?
+                  {
+                    'runs' => []
+                  }
+                else
+                  JSON.parse(shell_return.stdout)
```
ghbren, line 41: Please wrap this in a `begin...rescue...` because `JSON.parse` may throw an exception.

lib/salus/scanners/gitleaks.rb
```diff
@@ -0,0 +1,109 @@
...
+                end
+
+        report_runs(sarif['runs'])
+      end
+    end
+
+    def should_run?
+      true # Always run on the provided folder
```
ghbren, line 49: What happens if you don't specify any configs for Gitleaks?
Have you tested this out with `docker build ... local_salus`, then `docker run ... local_salus`?

lib/salus/scanners/gitleaks.rb
```diff
@@ -0,0 +1,109 @@
...
+  class Gitleaks < Base
+    def run
+      Dir.chdir(@repository.path_to_repo) do
+        copts = config_options
+        shell_return = run_shell(
+          'gitleaks '\
+          '--path . '\
+          '--format sarif '\
+          '--report /dev/stdout '\
+          '--leaks-exit-code 0 ' +
```
ghbren, line 18: It seems standard to use 1 as the failure exit code?

lib/salus/scanners/gitleaks.rb
```diff
@@ -0,0 +1,109 @@
...
+module Salus::Scanners
+  class Gitleaks < Base
+    def run
+      Dir.chdir(@repository.path_to_repo) do
+        copts = config_options
+        shell_return = run_shell(
+          'gitleaks '\
+          '--path . '\
+          '--format sarif '\
+          '--report /dev/stdout '\
```
ghbren, line 17: Is `--report` a required config option?

docs/scanners/gitleaks.md
ghbren: It doesn't look like gitleaks has an option to exclude dirs, correct?

lib/salus/scanners/gitleaks.rb
ghbren: and call `bugsnag_notify` if exception.

lib/salus/scanners/gitleaks.rb
ghbren: It appears 0 is instead normally used for success.

lib/salus/scanners/gitleaks.rb
kiwiz: `--report` is required to output to `stdout`. Otherwise, only debug data is output (to `stderr`).

kiwiz:
> There are 10+ binary files in the PR. Are all of them needed to test the commit scanning feature?

I haven't exhaustively tested them all, but I confirmed that removing a few will break tests.

lib/salus/scanners/gitleaks.rb
kiwiz: I'm using `--leaks-exit-code 0` to distinguish between a configuration error with gitleaks and results (which get processed further down).

PR #265, opened 2021-01-12
Update semgrep to version 0.36.0

kiwiz:
There are some changes in newer versions of semgrep, hence the large number of changes:

- The `semgrep-core` directory in the tgz no longer contains the Python module - just the `spacegrep` and `semgrep-core` binaries
- The errors now show the absolute path of the problematic file
- Parsing support has improved/changed - this resulted in different line numbers/columns, but since the intent of the target files is to be unparsable, I think this is fine.

These changes are on top of #257, but let me know if you'd like it on its own.


kiwiz:
@nishils Hi! Sorry for the direct ping - just wondering if you've got capacity to look at some of these PRs. I've got some other items I want to work on, but it helps to have less workstreams open (so I don't have to handle interactions between branches).

nishils (approved):

nishils:
Thanks for the ping actually! I get notifications much faster that way :)

PR #269, opened 2021-01-25
Use Docker multistage builds

kiwiz:
This PR updates the Dockerfile to take advantage of [multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/).

Highlights:
- Most of the dependencies are built in the `builder` image and copied to the final image
- Image size drops from 3.87GB to 1.22GB
- Light testing done (confirmed that all tests pass)


Dockerfile
```diff
@@ -110,29 +80,83 @@ ENV GO111MODULE on
...
+RUN pip3 install --user --no-cache-dir semgrep==${SEMGREP_VERSION}
+
+
+### Ruby
+COPY Gemfile Gemfile.lock ./
+RUN bundle install --deployment --without development:test
+
+
+
+FROM ruby:2.7.2-slim
```
nishils, line 109: Should we make 2.7.2 a docker environment variable here so that the top level image and this slim image are the same ruby versions?

nishils: Also, can you add in the SHA hash here like we have at the top?

kiwiz: Yup - should I update the hash to get the newest version of the image?

nishils: I think leave the one up top as is for now. We can go back and update when we do our regular patching and testing. Latest docker SHA commit for the slim image should be fine.

kiwiz: Ah, it seems that you can't parameterize the `FROM` directive like that.

nishils: Thanks for trying :(

nishils (dismissed):

ghbren:
@kiwiz Hi, how did you get the sha for ruby-slim-2.7.2?
I looked it up on dockerhub and saw a different sha.

kiwiz:
@ghbren I used the sha that showed up right after I pulled down the image locally. If you'd like, I can update the sha (or feel free to do so if the PR is updatable).

ghbren:
@kiwiz Which OS/Arch was the ruby-2.7.2-slim sha supposed to be for?
Thanks

kiwiz:
@ghbren Should be x86-64/Linux. It's possible I specified the wrong hash per:
```
$ docker image inspect ruby:2.7.2-slim
[
    {
        "Id": "sha256:4b1edb889e6cbb59c601745ce6a5a70514e63dbb37acdd4d7277e8615e9cffb0",
        "RepoTags": [
            "ruby:2.7.2-slim"
        ],
        "RepoDigests": [
            "ruby@sha256:b9eebc5a6956f1def4698fac0930e7a1398a50c4198313fe87af0402cab8d149"
        ],
```

ghbren:
@kiwiz Did you get the docker image here https://hub.docker.com/layers/ruby/library/ruby/2.7.2-slim/images/sha256-9748ccf1da3ce59b1a375a76b8318d97a6bb9322c1004600c78cfc41149f3431?context=explore?
I don't see x86-64/Linux listed.

kiwiz:
My bad, should be `linux/amd64`. I can't find that tag either. :thinking: Will take a closer look tomorrow.

ghbren:
> My bad, should be `linux/amd64`. I can't find that tag either. 🤔 Will take a closer look tomorrow.

Thanks. Never mind. I was just able to verify the sha.

ghbren:
@kiwiz Could you merge with master?  Then I'll approve it.
Thanks for the PR.

kiwiz:
@ghbren All set 👍

ghbren (approved):

PR #279, opened 2021-01-28
Add Default Sarif report

Kamsiy:
This PR adds the default sarif report format for unsupported scanners (scanners that currently do not support sarif)


lib/sarif/base_sarif.rb
```diff
@@ -0,0 +1,42 @@
...
+    DEFAULT_URI = "https://github.com/coinbase/salus".freeze
+  
+    attr_reader :name, :tool, :results, :scan_report
+    def initialize(scan_report)
+      @name = scan_report.scanner_name
+      @scan_report = scan_report
+      @tool = get_tool
+      @results = get_results
+    end
+
```
joshua-ostrom, line 12: SARIF Reference: http://docs.oasis-open.org/sarif/sarif/v2.0/csprd01/sarif-v2.0-csprd01.html

Great start.  We will want to build out the generic SARIF a bit further here.


Let's build out the 3.13.1 invocation object.

- 3.13.9 exitCodeDescription 
- 3.13.10 exitSignalName property
- 3.13.11 exitSignalNumber property
- 3.13.13 toolExecutionSuccessful property
Sample from the SARIF spec:
```
{
  "exitCode": 1,
  "exitCodeDescription": "Scan successful; warnings detected.",
  "toolExecutionSuccessful": true
}
```
I'd read through that section of the spec and see if which of these and other's would be a good fit.  Happy to pair on that too if you'd like.  The individual scanning results can follow but we will want to build out to results a little more for our generic SARIF here 👍 

**3.13.20 toolNotification**

```          "toolNotifications": [

            {
              "id": "SAL0001",
              "level": "note",
              "message": {
                "text": "Salus does yet support SARIF details for this scanner"
              }

            }
```
This will give us flexibility to display genuine scan CVE's / issues separately from meta about scanner support.

3.13.21 configurationNotifications could be an option too but I think 3.13.20 toolNotification better reflects the spirit of the SARIF spec here.

joshua-ostrom: We probably will a list of toolNotifications that Salus will support.  You can start that list with a single entry if you want as shown above:
```
SALUS_TOOL_NOTIFIICATIONS = { sal0001: "Salus does yet support SARIF details for this scanner" } 
```
or similar

lib/sarif/base_sarif.rb
```diff
@@ -0,0 +1,42 @@
...
+        "driver": {
+          "name" => @scan_report.scanner_name,
+          "semanticVersion" => @scan_report.version || "",
+          "informationUri" => DEFAULT_URI
+        }
+      }
+    end
+    
+    # Retreives result section for sarif report
+    def get_results
```
joshua-ostrom, line 25: See note above this can be an empty array for now

lib/sarif/base_sarif.rb
joshua-ostrom: Note too as you stand this up, if you find the implementation growing to full construct parts of the SARIF json you can favor composition over inheritance and spin up small dedicated classes for each section. One for toolNotification for example.   Not a have to have, just a note to help you keep the code testable / manageable / SOLID.

spec/lib/sarif/base_sarif_spec.rb
```diff
@@ -0,0 +1,35 @@
+require_relative '../../spec_helper.rb'
```
joshua-ostrom, line 1: 💯 🎉 for specs up front here !

spec/lib/sarif/base_sarif_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+describe Sarif::BaseSarif do
```
joshua-ostrom, line 3: One thought as you layout all of the specs.  It would be really nice to include validation against the SARIF json schema in the specs.  That way we know as we build out the SARIF that it conforms to the spec.

See https://github.com/ruby-json-schema/json-schema

We could add the schema as a fixture here and validate against it.  That will ensure this first pass is valid and as you stand up scanners we stay compliant.  

```
JSON::Validator.validate(schema, your_generated_sarif_here)
```

joshua-ostrom (commented):
Awesome start!  🎉 

Provided some comments along the way.

lib/sarif/base_sarif.rb
```diff
@@ -0,0 +1,42 @@
...
+      @results = results_info
+    end
+
+    # Retrieve tool section for sarif report
+    def tool_info
+      {
+        "driver": {
+          "name" => @scan_report.scanner_name,
+          "semanticVersion" => @scan_report.version || "",
+          "informationUri" => DEFAULT_URI
```
kiwiz, line 19: - `semanticVersion` might not apply to all the scanners, so it might be better to use `version` here
-  `DEFAULT_URI` is a better fit for `run.conversion.tool.driver`

lib/sarif/base_sarif.rb
Kamsiy: The toolNotification object is currently available in sarif version 2.0. 

In [sarif version 2.1 ](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317588) there's a toolExecutionNotifications property (3.20.21) which has a similar function.

lib/sarif/base_sarif.rb
```diff
@@ -0,0 +1,68 @@
...
+      @tool = tool_info
+      @results = results_info
+    end
+
+    # Retrieve tool section for sarif report
+    def tool_info
+      {
+        "driver": {
+          "name" => scan_report.scanner_name,
+          "version" => scan_report.version || ""
```
ghbren, line 19: You don't need to `|| ""` because the scanner super class sets `version` to empty string.

lib/sarif/sarif_report.rb
```diff
@@ -0,0 +1,42 @@
...
+      @scan_reports.each { |scan_report| sarif_report["runs"] << converter(scan_report[0]) }
+      JSON.pretty_generate(sarif_report)
+    end
+
+    # Converts a ScanReport to a sarif report for the given scanner
+    #
+    # @params sarif_report [Salus::ScanReport]
+    # @return
+    def converter(scan_report)
+      adapter = "Sarif::#{scan_report.scanner_name.capitalize}Sarif"
```
ghbren, line 34: No need to `capitalize`.

spec/lib/sarif/sarif_report_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+      custom_info = { bitcoin_price: 100_000 }
+      config = { lemurs: 'contained', raptors: 'loose' }
+      report = Salus::Report.new(project_name: name, custom_info: custom_info, config: config)
+      scan_reports.each { |scan_report| report.add_scan_report(scan_report, required: false) }
+      schema = JSON.parse(File.read('spec/fixtures/sarif/sarif-schema.json'))
+
+      body = JSON.parse(report.to_sarif)
+      report1 = body['runs'][0]['tool']
+      report2 = body['runs'][1]['tool']
+      expect(JSON::Validator.validate(schema, body)).to be true
```
ghbren, line 24: Can you update the base sarif class to always run `validate`?

lib/sarif/sarif_report.rb
```diff
@@ -0,0 +1,42 @@
...
+      converter = Object.const_get(adapter).new(scan_report)
+      converter.sarif_report
+    rescue NameError
```
ghbren, line 37: If I understand your intention correctly, `NameError` is only supposed to associate with `Object.const_get`.
But why is `converter.sarif_report` also part of the `begin ...` block?

spec/lib/sarif/sarif_report_spec.rb
ghbren: I would then include tests cases showing what happens when the sarif report is valid / invalid.

lib/sarif/sarif_report.rb
Kamsiy: The goal was to have the converter.sarif_report returned in either situation (If an adapter exists or if an adapter does not exist). Is there a cleaner way to achieve this? since I am repeating the same thing twice?

ghbren: I would move `converter.sarif_report` out of the the `begin...` block.

You can just return empty sarif report if there is an NameError, like
```ruby
begin ...
rescue NameError
  # return empty sarif report
end

converter.sarif_report
```

Now, a `return` inside `rescue` should work (unless you have an `ensure` block too).  But I'm not 100% sure.  Please test it.

Kamsiy: I tried an ensure block but it does not return anything. 

The rescue block is run if the adapter for a scanner cannot be instantiated. It would run the sarif_report for the BaseSarif class. However, if the converter/adapter can be instantiated, then it would run the sarif_report for that scanner's adapter (that is it would run GosecSarif.sarif_report instead of BaseSarif.sarif_report). 

I tried your suggestion but I had an issue. Having the function call in the rescue block alone would mean an instantiated adapter would not return a sarif_report. Because of this i call the sarif_report on both blocks(begin and rescue) because its called on two different objects

lib/sarif/sarif_report.rb
```diff
@@ -0,0 +1,55 @@
...
+    #
+    # @params sarif_report [Salus::ScanReport]
+    # @return
+    def converter(scan_report)
+      adapter = "Sarif::#{scan_report.scanner_name}Sarif"
+      begin
+        converter = Object.const_get(adapter).new(scan_report)
+      rescue NameError
+        converter = BaseSarif.new(scan_report)
+        return converter.build_runs_object
```
joshua-ostrom, line 50: is this return redundant?

lib/sarif/sarif_report.rb
```diff
@@ -0,0 +1,55 @@
...
+require 'json-schema'
+require_relative './base_sarif'
+module Sarif
+  # Class for generating sarif reports
+  class SarifReport
+    class SarifInvalidFormatError < StandardError; end
+
+    SARIF_VERSION = "2.1.0".freeze
+
+    SARIF_SCHEMA = "https://schemastore.azurewebsites.net/schemas"\
```
joshua-ostrom, line 11: Any reason for this site vs the SARIF site itself for sourcing the schema? 
http://docs.oasis-open.org/sarif/sarif/v2.0/csprd01/schemas/sarif-schema.json

ghbren (approved):

PR #280, opened 2021-01-29
Disable semgrep version check

kiwiz:
This causes the error handling [logic](https://github.com/kiwiz/salus/blob/0374056d0c049626b2f1e9e92a08f07582628c6b/lib/salus/scanners/semgrep.rb#L133) to pick up the wrong line if a new version exists.


ghbren:
could you merge with master?

kiwiz:
@ghbren 👍

PR #286, opened 2021-02-08
Add SARIF reporting for gosec

Kamsiy:
This PR adds sarif reporting options for gosec


lib/sarif/sarif_report.rb
```diff
@@ -0,0 +1,56 @@
...
+        "$schema" => SARIF_SCHEMA,
+        "runs" => []
+      }
+      # for each scanner report, run the appropriate converter
+      @scan_reports.each { |scan_report| sarif_report["runs"] << converter(scan_report[0]) }
+      report = JSON.pretty_generate(sarif_report)
+      path = File.expand_path('schema/sarif-schema.json', __dir__)
+      schema = JSON.parse(File.read(path))
+
+      if JSON::Validator.validate(schema, report)
```
egonz, line 35: Is it worth indicating how validation failed? E.g.

`#`
`# return an array of error messages when validation fails`
`#`
`JSON::Validator.fully_validate(schema, { "a" => "taco" })`

lib/sarif/sarif_report.rb
```diff
@@ -0,0 +1,56 @@
...
+    # Converts a ScanReport to a sarif report for the given scanner
+    #
+    # @params sarif_report [Salus::ScanReport]
+    # @return
+    def converter(scan_report)
+      adapter = "Sarif::#{scan_report.scanner_name}Sarif"
+      begin
+        converter = Object.const_get(adapter).new(scan_report)
+      rescue NameError
+        converter = BaseSarif.new(scan_report)
```
egonz, line 50: Can you clarify what this is doing?

egonz (commented):
Great work on this. Left a few questions.

lib/sarif/sarif_report.rb
Kamsiy: It tries to create an adapter using the scanners name. If successful, the scanners adapter is instantiated and then run. However, if theres a NameError(could not be instantiated) then the BaseSarif(the default sarif reporter) class is intatiated

lib/sarif/sarif_report.rb
Kamsiy: Yes, information about the error will be helpful. I've added some documentation for this

ghbren:
For each vulnerability, gosec outputs severity and confidence. Can you incorporate these to sarif?

lib/sarif/gosec_sarif.rb
```diff
@@ -0,0 +1,81 @@
...
+      rescue JSON::ParserError
+        @logs = []
+      end
+    end
+
+    def parse_error(error)
+      line = error['line'].to_i
+      column = error['column'].to_i
+      line = 1 if line.zero?
+      column = 1 if column.zero?
```
ghbren, line 29: Why set line and column to 1 if they are 0?
I'd keep them 0 to be consistent with the Gosec ouptut.

lib/sarif/base_sarif.rb
```diff
@@ -0,0 +1,146 @@
...
+          "driver": {
+            "name": "Salus",
+            "informationUri": DEFAULT_URI
+          }
+        }
+      }
+    end
+
+    # Returns a sarif wraning level for a given severity
+    def sarif_level(severity)
```
ghbren, line 133: How did you decide on the severity to level mappings?
Also, we need separate keys for severity and confidence.

ghbren:
vulnerability column numbers are missing, even though line numbers are included.

lib/sarif/gosec_sarif.rb
```diff
@@ -0,0 +1,81 @@
...
+    def initialize(scan_report)
+      super(scan_report)
+      @uri = GOSEC_URI
+      begin
+        json_obj = JSON.parse(scan_report.log(''))
+        issues = json_obj['Issues']
+        errors = json_obj['Golang errors']
+        parsed_errors = []
+        errors.each do |key|
+          key[1].each do |location|
```
ghbren, line 14: You can combine the two lines above to something like
```ruby
errors.each {|key, location|
```

lib/sarif/gosec_sarif.rb
```diff
@@ -0,0 +1,81 @@
...
+            location['uri'] = key[0]
+            parsed_errors << parse_error(location)
+          end
+        end
+        @logs = issues.concat parsed_errors
+      rescue JSON::ParserError
+        @logs = []
```
ghbren, line 21: In this case, I think you should output an error.  Maybe have `to_sarif` raise an exception?

lib/sarif/base_sarif.rb
Kamsiy: In the [SARIF document](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.pdf), there are defined fields for different objects. For the result object(@3.27), there was no mention about confidence or severity. The closest option for these two is the level object(@3.27.10). If i add severity and confidence fields the program would raise a SarifInvalidFormatError because it would not follow the schema. The mappings were done based on how gosec currently generates its levels from its severity. https://github.com/securego/gosec/blob/master/output/sarif_format.go#L173

lib/sarif/gosec_sarif.rb
Kamsiy: This was because in a file the start position is row 1, column 1. Gosec sometimes returns row 0, column 0 because it couldn't find an exact location for items in "Golang Errors" and this would cause the program to throw an SarifInvalidFormatError. To avoid this i made row 1 column 1, the default position

lib/sarif/gosec_sarif.rb
Kamsiy: Its not an error that should lead to failure. If a ParseError is raised, it could mean that there was nothing logged by gosec for that repo because it either generated an error or it passed successfully. If theres an error, its handled in the build_invocations method

lib/sarif/gosec_sarif.rb
ghbren: Sounds good1

ghbren (approved):

PR #288, opened 2021-02-10
update brakeman from 5.0.0 to older version 4.10.0

ghbren:
Brakeman 5.0.0 was scanning files/dirs that it was not scanning before


egonz (approved):
Looks good

PR #290, opened 2021-02-10
version bump to 2.10.16

ghbren:
Downgraded brakeman from 5.0.0 to 4.10.0. We plan to do more work before deploying  brakeman 5.0.0.


egonz (approved):
Looks good.

PR #310, opened 2021-02-24
Handle duplicate errors and issues in gosec report

Kamsiy:
This PR fixes salus error for duplicate entries in SARIF results object for gosec

### DIFF
--- a/lib/salus/report.rb
+++ b/lib/salus/report.rb
@@ -1,10 +1,12 @@
 require 'faraday'
 require 'json'
 require 'salus/formatting'
+require 'salus/bugsnag'
 
 module Salus
   class Report
     include Formatting
+    include Salus::SalusBugsnag
 
     class ExportReportError < StandardError; end
 
@@ -125,7 +127,7 @@ module Salus
     def to_sarif
       Sarif::SarifReport.new(@scan_reports).to_sarif
     rescue Sarif::SarifReport::SarifInvalidFormatError => e
-      e.message
+      bugsnag_notify(e.message)
     end
 
     # Send the report to given URIs (which could be remove or local).
--- a/lib/salus/report.rb
+++ b/lib/salus/report.rb
@@ -127,7 +127,7 @@ module Salus
     def to_sarif
       Sarif::SarifReport.new(@scan_reports).to_sarif
     rescue Sarif::SarifReport::SarifInvalidFormatError => e
-      bugsnag_notify(e.message)
+      bugsnag_notify(e.message + "\nBuild Info:" + @builds.to_s)
     end
 
     # Send the report to given URIs (which could be remove or local).
--- a/lib/salus/report.rb
+++ b/lib/salus/report.rb
@@ -126,8 +126,8 @@ module Salus
 
     def to_sarif
       Sarif::SarifReport.new(@scan_reports).to_sarif
-    rescue Sarif::SarifReport::SarifInvalidFormatError => e
-      bugsnag_notify(e.message + "\nBuild Info:" + @builds.to_s)
+    rescue StandardError => e
+      bugsnag_notify(e.class.to_s + " " + e.message + "\nBuild Info:" + @builds.to_s)
     end
 
     # Send the report to given URIs (which could be remove or local).
--- a/lib/salus/report.rb
+++ b/lib/salus/report.rb
@@ -124,6 +124,8 @@ module Salus
 
     def to_sarif
       Sarif::SarifReport.new(@scan_reports).to_sarif
+    rescue Sarif::SarifReport::SarifInvalidFormatError => e
+      e.message
     end
 
     # Send the report to given URIs (which could be remove or local).
--- a/lib/sarif/gosec_sarif.rb
+++ b/lib/sarif/gosec_sarif.rb
@@ -1,5 +1,9 @@
+require 'salus/bugsnag'
+
 module Sarif
   class GosecSarif < BaseSarif
+    include Salus::SalusBugsnag
+
     GOSEC_URI = 'https://github.com/securego/gosec'.freeze
 
     def initialize(scan_report)
@@ -22,7 +26,8 @@ module Sarif
       end
       parsed_errors.compact!
       issues.concat parsed_errors
-    rescue JSON::ParserError
+    rescue JSON::ParserError => e
+      bugsnag_notify(e.message)
       []
     end
 
--- a/lib/sarif/gosec_sarif.rb
+++ b/lib/sarif/gosec_sarif.rb
@@ -53,7 +53,7 @@ module Sarif
       if issue[:id] == 'SAL0002'
         issue
       else
-        id = issue['details'] + issue['file'] + issue['line']
+        id = issue['details'] + ' ' + issue['file'] + ' ' + issue['line']
         return nil if @issues.include?(id)
 
         @issues.add(id)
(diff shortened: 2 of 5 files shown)


lib/salus/report.rb
```diff
@@ -124,6 +124,8 @@ def to_json
 
     def to_sarif
       Sarif::SarifReport.new(@scan_reports).to_sarif
+    rescue Sarif::SarifReport::SarifInvalidFormatError => e
+      e.message
```
joshua-ostrom, line 128: You'll want a bugsnag_notify here

lib/sarif/gosec_sarif.rb
```diff
@@ -5,6 +5,7 @@ class GosecSarif < BaseSarif
     def initialize(scan_report)
       super(scan_report)
       @uri = GOSEC_URI
+      @issues = Set.new
```
joshua-ostrom, line 8: Minor nit... I think you'd have a small readability win if you move the parsing of the scan report into its own method.

That would thin out the initialize method to something like

```
def initialize(scan_report)
  super(scan_report)
  @uri = GOSEC_URI
  parse_scan_report!(scan_report)
end
```

spec/lib/sarif/gosec_sarif_spec.rb
```diff
@@ -7,6 +7,51 @@
...
+        end
+      end
+    end
+
+    describe '#sarif_level' do
+      context 'gosec severities' do
+        let(:repo) { Salus::Repo.new('spec/fixtures/gosec/safe_goapp') }
+        it 'are mapped to sarif levels' do
+          scan_report = Salus::ScanReport.new(scanner_name: "Gosec")
+          adapter = Sarif::GosecSarif.new(scan_report)
```
joshua-ostrom, line 47: 💯  Thanks for adding coverage on the expected mappings

lib/salus/report.rb
```diff
@@ -124,6 +126,8 @@ def to_json
 
     def to_sarif
       Sarif::SarifReport.new(@scan_reports).to_sarif
+    rescue Sarif::SarifReport::SarifInvalidFormatError => e
+      bugsnag_notify(e.message)
```
ghbren, line 130: It would help to add project build info (which includes like buildkite url) to `bugsnag_notify`.  Otherwise, we won't know which repo produced the exception, which may make it hard to debug.

You may be able to get `build` info with `@config.builds`, like in processor.rb.  Then you can just use `@config.builds.to_s`, or set it to empty string if value is `nil`.

lib/sarif/gosec_sarif.rb
```diff
@@ -5,28 +5,37 @@ class GosecSarif < BaseSarif
...
+          parsed_errors << parse_error(location)
         end
-        @logs = issues.concat parsed_errors
-      rescue JSON::ParserError
-        @logs = []
       end
+      parsed_errors.compact!
+      issues.concat parsed_errors
+    rescue JSON::ParserError
+      []
```
ghbren, line 26: Should we also bugsnag notify here?

ghbren: This is a good case where adding build info to the bugsnag message would be helpful.

lib/sarif/gosec_sarif.rb
```diff
@@ -5,28 +5,37 @@ class GosecSarif < BaseSarif
...
     end
 
     def parse_error(error)
       line = error['line'].to_i
       column = error['column'].to_i
       line = 1 if line.zero?
       column = 1 if column.zero?
+
+      id = error['error'] + error['uri'] + line.to_s
```
ghbren, line 35: Well, this could cause issues with edge cases.

Let's say
error 1 has uri "http://abcd.com/1", line 123
error 2 has uri "http://abcd.com/11", line 23

The code will consider these two different issues equivalent.

I think a simple fix is just update the code to 
`id = error['error'] + ' ' + error['uri'] + ' ' + line.to_s`

lib/sarif/gosec_sarif.rb
Kamsiy: scan reports do not have configs, i would have to pass it down from reports. report -> scan_report -> gosec_sarif

lib/salus/report.rb
```diff
@@ -124,6 +126,8 @@ def to_json
 
     def to_sarif
       Sarif::SarifReport.new(@scan_reports).to_sarif
+    rescue Sarif::SarifReport::SarifInvalidFormatError => e
+      bugsnag_notify(e.message + "\nBuild Info:" + @builds.to_s)
```
joshua-ostrom, line 130: 💯

lib/sarif/gosec_sarif.rb
```diff
@@ -5,28 +5,37 @@ class GosecSarif < BaseSarif
...
-        issues = json_obj['Issues']
-        errors = json_obj['Golang errors']
-        parsed_errors = []
-        errors.each do |key|
-          key[1].each do |location|
-            location['uri'] = key[0]
-            parsed_errors << parse_error(location)
-          end
+      @issues = Set.new
+      @logs = parse_scan_report!(scan_report)
```
joshua-ostrom, line 9: 💯 much leaner initialize

lib/sarif/gosec_sarif.rb
joshua-ostrom, line 26: what do you think about a bugsnag log here too?  Any time we swallow an exception we should have some record of it.

spec/lib/sarif/gosec_sarif_spec.rb
```diff
@@ -7,6 +7,51 @@
...
+          expect(adapter.sarif_level("MEDIUM")).to eq("error")
```
joshua-ostrom, line 48: thanks for adding explicit checks here

lib/sarif/gosec_sarif.rb
Kamsiy: I've added a bugsnag log

lib/salus/report.rb
```diff
@@ -124,6 +126,8 @@ def to_json
 
     def to_sarif
       Sarif::SarifReport.new(@scan_reports).to_sarif
+    rescue Sarif::SarifReport::SarifInvalidFormatError => e
```
ghbren, line 129: Could there be any other types of exceptions from `Sarif::SarifReport.new(@scan_reports).to_sarif`?

lib/sarif/gosec_sarif.rb
```diff
@@ -1,32 +1,46 @@
...
+      []
     end
 
     def parse_error(error)
       line = error['line'].to_i
       column = error['column'].to_i
       line = 1 if line.zero?
       column = 1 if column.zero?
+
+      id = error['error'] + error['uri'] + line.to_s
```
ghbren, line 40: I added a comment about how the `id = ...` could be buggy.   The comment was marked as resolved, but I don't see  any changes to the line?

ghbren, line 40: `id = error['error'] + error['uri'] + line.to_s` should be updated to  `id = error['error'] + ' ' + error['uri'] + ' ' + line.to_s`.

kiwiz:
Might be fixed as of https://github.com/securego/gosec/pull/565

ghbren:
> Might be fixed as of [securego/gosec#565](https://github.com/securego/gosec/pull/565)

This fix above is for how we check for duplicates, but thanks for comment @kiwiz .

ghbren (approved):

PR #318, opened 2021-03-09
Add request parameter configuration for salus

Kamsiy:
This Pr adds parameter configurations for salus. Users can specify request parameters in config files


docs/configuration.md
```diff
@@ -59,6 +62,19 @@ reports:
...
+    params:
+      repo: 'Random Repo'
+      user: 'John Doe' 
+  - uri: https://salus-config.internal2.net/salus-report
+    format: json
+    verbose: true
+    params:
+      repo: 'Random Repo'
+      user: 'John Doe' 
+      report: null
```
joshua-ostrom, line 77: On small tweak here.  To prevent a hard coded param name for the report itself we could just create a dedicated yaml config for that and then params could represent any additional ad hoc params to include in the POST

```
    salus_report_param_name: report
    additional_params:
      repo: 'Random Repo'
      user: 'John Doe' 
```

joshua-ostrom: but I'm not married to the names there, just the idea that we let the user decide the field names including the report one 👍

docs/configuration.md
```diff
@@ -59,6 +62,19 @@ reports:
   - uri: https://salus-config.internal.net/salus-report
     format: json
     verbose: true
+  - uri: https://salus-config.internal2.net/salus-report
+    format: json
+    verbose: true
```
joshua-ostrom, line 67: These will need to be updated for the final param names we landed on

lib/salus/report.rb
```diff
@@ -214,6 +212,28 @@ def send_report(remote_uri, data, format)
...
+      body_hash
+    end
+
+    def report_body(config)
+      return report_body_hash(config, to_s(verbose: verbose)).to_s if config['format'] == 'txt'
+
+      body = report_body_hash(config, JSON.parse(to_json)) if config['format'] == 'json'
+      body = report_body_hash(config, JSON.parse(to_sarif)) if config['format'] == 'sarif'
+      return JSON.pretty_generate(body) if config['format'] == 'json'
+
```
joshua-ostrom, line 231: Should this be a if ['json','sarif'].includes?(config['format'])

docs/configuration.md
```diff
@@ -51,14 +51,30 @@ custom_info: "PR-123"
...
+    post:
+      salus_report_param_name: 'report'
+      additional_params:
+        repo: 'Random Repo'
+        user: 'John Doe' 
+        report: null
```
joshua-ostrom, line 77: report:null can be dropped

spec/lib/salus/report_spec.rb
```diff
@@ -178,6 +178,35 @@ def build_report(report_uri)
       end
     end
 
+    context 'HTTP report URI given with request parameters' do
+      it 'should make a call to send the report for http URI' do
```
joshua-ostrom, line 182: This spec looks good.  It's testing format: json. We should probably copy and make a test for txt, yaml, sarif too

spec/lib/salus/report_spec.rb
```diff
@@ -178,6 +178,123 @@ def build_report(report_uri)
...
+            'User-Agent' => 'Faraday v1.3.0',
+            'X-Scanner' => 'salus' }).to_return(status: 200, body: "", headers: {})
+
+        expect { report.export_report }.not_to raise_error
+      end
+      it 'should make a call to send the sarif report for http URI' do
+        url = 'https://nerv.tk3/salus-report'
+        params = { 'salus_report_param_name' => 'report',
+          'additional_params' => { "foo" => "bar", "abc" => "def" } }
+        directive = { 'uri' => url, 'format' => 'yaml', 'post' => params, 'verbose': false }
```
joshua-ostrom, line 274: Any reason not to set the format to sarif on line 274 instead of line 276?

spec/lib/salus/report_spec.rb
```diff
@@ -178,6 +178,123 @@ def build_report(report_uri)
...
+        expect { report.export_report }.not_to raise_error
+      end
+
+      it 'should make a call to send the txt report for http URI' do
+        url = 'https://nerv.tk3/salus-report'
+        params = { 'salus_report_param_name' => 'report',
+          'additional_params' => { "foo" => "bar", "abc" => "def" } }
+        directive = { 'uri' => url, 'format' => 'yaml', 'post' => params, 'verbose': false }
+        build_report(directive)
+        directive['format'] = 'txt'
```
joshua-ostrom, line 237: Any reason not to set the format to txt on line 235 instead of line 237?

spec/lib/salus/report_spec.rb
Kamsiy: I was only trying to avoid modifying other tests. the scan_report used in build directive does not have a version and contains duplicate reports. generating a sarif report from this scan report would return an empty report because it would generate an invalid sarif report.
https://github.com/coinbase/salus/blob/02f7678bca84c8c89d03a36059a03373ec746fad/spec/lib/salus/report_spec.rb#L125-L142

Kamsiy: i was trying to find a work around to avoid using the scan reports from build_reports, the build_reports has duplicate reports. Sarif returns an empty report when there’s duplicate entries

joshua-ostrom (approved):

PR #324, opened 2021-03-17
support for --filter_sarif

ghbren:


joshua-ostrom (approved):
Looks good, we may want to expand to filter non sarif output too in the future, but this change looks good 👍

PR #328, opened 2021-03-18
Add suppressions for unenforced scanners

Kamsiy:
This PR adds suppression objects to results of unenforced scanners. This would allow users the ability to include unenforced or allow only enforced scanners in their results using the include_suppressed configuration

### DIFF
--- a/lib/sarif/base_sarif.rb
+++ b/lib/sarif/base_sarif.rb
@@ -112,7 +112,6 @@ module Sarif
       end
 
       invocation = build_invocations(@scan_report, supported)
-      invocation[:executionSuccessful] = true if @required == false
       {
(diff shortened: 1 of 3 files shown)


lib/sarif/base_sarif.rb
```diff
@@ -96,25 +96,28 @@ def build_runs_object(supported)
         parsed_issue = parse_issue(issue)
         next if !parsed_issue
         next if parsed_issue[:suppressed] && @config['include_suppressed'] == false
+        next if @required == false && @config['include_suppressed'] == false
```
joshua-ostrom, line 99: 👍

joshua-ostrom (approved):
LGTM

PR #329, opened 2021-03-19
add builds info to sarif diff reports

ghbren:
Update to include project build info in --filter_sarif reports


joshua-ostrom (approved):
LGTM

PR #332, opened 2021-03-24
add misses section to Semgrep and PatternSearch json

ghbren:
PatternSearch / Semgrep json contained structured `hits` info for patterns that were detected, like below.  But there is no structured section for patterns that were required but not detected.
```json
        "hits": [
          {
            "regex": "bar",
            "forbidden": true,
            "required": false,
            "msg": "bar is forbidden",
            "hit": "doh.js:1:bar"
          },
```

This PR adds the "misses" section, like
```json
        "misses": [
          {
            "regex": "foo",
            "forbidden": false,
            "required": true,
            "msg": "foo is required"
          },
```

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -35,7 +35,7 @@ Metrics/ClassLength:
   Max: 1000
 
 Metrics/AbcSize:
-  Max: 100
+  Max: 150
 
 Metrics/CyclomaticComplexity:
   Max: 30
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -35,7 +35,7 @@ Metrics/ClassLength:
   Max: 1000
 
 Metrics/AbcSize:
-  Max: 150
+  Max: 100
 
 Metrics/CyclomaticComplexity:
   Max: 30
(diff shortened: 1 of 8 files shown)


.rubocop.yml
```diff
@@ -35,7 +35,7 @@ Metrics/ClassLength:
   Max: 1000
 
 Metrics/AbcSize:
-  Max: 100
+  Max: 150
```
joshua-ostrom, line 38: 150 is really high for AbcSize (IIRC default is 15).  If we have a method that's violating we should probably fix or disable the check for that method via # rubocop:disable Metrics/AbcSize.

lib/salus/scanners/pattern_search.rb
```diff
@@ -96,6 +97,12 @@ def run
               if match['required']
                 failure_messages << "Required pattern \"#{match['regex']}\" was not found " \
                   "- #{match['message']}"
+                all_misses << {
```
joshua-ostrom, line 100: This looks good.  Question on the all hits block above... I see that we are passing sift the -n (show line numbers).  As we update for SARIF we could consider including a field on the hits for the line of code information (or two to include the filename too).  Thoughts?

joshua-ostrom (approved):

PR #339, opened 2021-03-29
fix pattern search sarif

ghbren:
Fix bugs in PatternSearch sarif issues
* In `parse_miss`, deleted `return nil if miss[:msg].nil?` because the value can never be `nil`.  It could be empty string though, which `message()` already takes care of.
* In `parse_miss`, deleted `return nil if miss[:msg].include?("Required")` because the user-configurable `:msg` is irrelevant here.
* In `parse_miss` and `parse_hit`, removed unneeded (and buggy) code to eliminate duplicated vulns. 

Tested on actual repo with various PatternSearch rules.

### DIFF
--- a/lib/sarif/pattern_search_sarif.rb
+++ b/lib/sarif/pattern_search_sarif.rb
@@ -56,12 +56,7 @@ module Sarif
     end
 
     def parse_miss(miss)
-      return nil if miss[:msg].nil?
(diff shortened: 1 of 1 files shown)


joshua-ostrom (approved):
Approved assuming we don't have dups coming from pattern search

PR #350, opened 2021-04-13
update project_name to not allow whitespaces or ;

ghbren:


joshua-ostrom (approved):
Looks good we can probably relax the space etc later but this looks good.

PR #358, opened 2021-05-06
add plugin manager

ghbren:


spec/lib/salus_spec.rb
```diff
@@ -172,5 +173,30 @@
         end
       end
     end
+
+    context 'With plugins' do
+      it 'Should update config based on plugin' do
+        Dir.chdir('spec/fixtures/blank_repository2') do
+          plugin_dir = "../../../lib/salus/plugins"
+          FileUtils.cp('test_plugins/my_filter.rb', plugin_dir + '/my_filter.rb')
```
joshua-ostrom, line 181: Thanks for including the spec 👍 

One thought here, if the logic for the directory was placed into a method you could do away with the file copying / deleting here and just stub the value:

`expect(Salus::PluginManager).to receive(:plugin_directory).and_return('spec/fixtures/blank_repository2/test_plugins')`

ghbren: updated accordingly

lib/salus/config.rb
```diff
@@ -55,6 +57,11 @@ def initialize(configuration_files = [], ignore_ids = [])
       # Check if any of the values are actually pointing to envars.
       final_config = fetch_envars(final_config)
 
+      Salus::PluginManager.load_plugins
```
joshua-ostrom, line 60: We should probably move this to lib/salus.rb.  That would ensure that the plugins are loaded and ready as we start enabling other hooks (beyond the config hook).

joshua-ostrom (approved):

PR #361, opened 2021-05-12
fix failed specs due to bundle audit CVE update

ghbren:
Bundle Audit updated the CVE on a package, which caused the specs to fail.


joshua-ostrom (approved):
Nice job in cleaning those up while you were in there

PR #362, opened 2021-05-12
add report filter

ghbren:
Added reporter filter support, so we can add something like
```ruby
module Filter
  module TestReportFilter
    def self.filter_report_hash(report_hash)
      report_hash[:config][:builds] ||= {}
      report_hash[:config][:builds][:mykey] = 'myval'
      report_hash
    end
  end

  Salus::Report.register_filter(Filter::TestReportFilter)
end
```

Then the report hash will include `[:config][:builds][:mykey] = 'myval'`

### DIFF
--- a/lib/salus/report.rb
+++ b/lib/salus/report.rb
@@ -25,6 +25,8 @@ module Salus
 
     attr_reader :builds
 
+    @@filters = []
+
     def initialize(report_uris: [], builds: {}, project_name: nil, custom_info: nil, config: nil,
                    repo_path: nil, filter_sarif: nil, ignore_config_id: nil)
       @report_uris = report_uris     # where we will send this report
@@ -38,6 +40,17 @@ module Salus
       @filter_sarif = filter_sarif   # Filter out results from this file
       @repo_path = repo_path         # path to repo
       @ignore_config_id = ignore_config_id # ignore id in salus config
+    end
+
+    def apply_report_hash_filters(report_hash)
+      @@filters.each do |filter|
+        report_hash = filter.filter_config(report_hash) if filter.respond_to?(:filter_config)
+      end
+      report_hash
+    end
+
+    def self.register_filter(filter)
+      @@filters << filter
     end
 
     def record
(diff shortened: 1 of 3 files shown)


lib/salus/report.rb
```diff
@@ -40,6 +42,17 @@ def initialize(report_uris: [], builds: {}, project_name: nil, custom_info: nil,
       @ignore_config_id = ignore_config_id # ignore id in salus config
     end
 
+    def apply_report_hash_filters(report_hash)
+      @@filters.each do |filter|
+        report_hash = filter.filter_config(report_hash) if filter.respond_to?(:filter_config)
```
joshua-ostrom, line 47: We will want to avoiding overriding filter_config (that hook is for filtering the config).  We should use filter_report_hash here:

```
def apply_report_hash_filters(report_hash)
  @@filters.each do |filter|
    report_hash = filter.filter_report_hash(report_hash) if filter.respond_to?(:filter_report_hash)
  end
  report_hash
end
```
That will ensure that we have different hooks for overriding configs ( filter_config ) and overriding reports hashes ( filter_report_hash )

spec/fixtures/blank_repository2/test_plugins/report_filter.rb
```diff
@@ -0,0 +1,11 @@
+module Filter
+  module TestReportConfig
+    def self.filter_config(report_hash)
```
joshua-ostrom, line 3: We should change this to filter_report_hash for good measure.  That will avoid any potential collisions with modules that want to support both hooks (config and report hash)

lib/salus/report.rb
ghbren: Thanks, updated accordingly.

spec/fixtures/blank_repository2/test_plugins/report_filter.rb
ghbren: Thanks, updated accordingly.

joshua-ostrom (approved):

PR #363, opened 2021-05-12
add supported_languages

ghbren:
added `def supported_languages` for some scanners to return an array of supported languages. 
Default will be `[]`.


joshua-ostrom (approved):
We may want to revisit the language agnostic ones (like PatternSearch) to return a wildcard in the future - but I think this looks perfect for our initial pass

PR #369, opened 2021-05-28
support running gosec from a configurable list of subdirs

ghbren:
Support option to run gosec only from specified subdirs.  Example salus.yaml
```yaml
scanner_configs:
  Gosec:
    run_from_dirs:
      - app1
      - app2
      - app3
```

### DIFF
--- a/lib/salus/scanners/gosec.rb
+++ b/lib/salus/scanners/gosec.rb
@@ -7,26 +7,77 @@ require 'json'
 module Salus::Scanners
   class Gosec < Base
     def run
+      return run_from_dir if @config['run_from_dirs'].nil?
+
+      @gosec_failed = false
+      @gosec_stderr = ''
+      @gosec_stdout = ''
+      @gosec_json = {}
+      run_from_dirs = @config['run_from_dirs'].sort
+      @config.delete('run_from_dirs')
+      run_from_dirs.each do |dir|
+        if dir.start_with?("/") || dir.start_with?("./") || dir.include?("..")
+          msg = "run_from_dirs values should be relative paths to subdirs in the repo " \
+                "and cannot start with ./"
+          report_stderr(msg)
+          report_error(
+            msg,
+            status: 1
+          )
+          return report_failure
+        else
+          run_from_dir(dir)
+        end
+      end
+
+      return report_success if @gosec_failed == false
+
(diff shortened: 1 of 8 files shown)


lib/salus/scanners/gosec.rb
```diff
@@ -7,26 +7,81 @@
...
+          run_from_dir(dir)
+        end
+      end
+
+      # gosec runs from multiple dirs
+      # success only if none of the runs set @gosec_failed = true
+      return report_success if @gosec_failed == false
+
+      log(JSON.pretty_generate(@gosec_json)) if !@gosec_json.empty?
+      report_stderr(@gosec_stderr) if @gosec_stderr.empty?
```
joshua-ostrom, line 40: should this if be an unless ?

lib/salus/scanners/gosec.rb
```diff
@@ -7,26 +7,81 @@
...
+      report_stdout(@gosec_stdout) if @gosec_stdout.empty?
```
joshua-ostrom, line 41: should this if be an unless?

lib/salus/scanners/gosec.rb
```diff
@@ -7,26 +7,81 @@
...
+      # 'run_from_dirs' specifies a list of subdirs to run salus from
+      # if not specified, then run_from_dir will mimic the original gosec 'run' behavior
+      return run_from_dir if @config['run_from_dirs'].nil?
+
+      @gosec_failed = false
+      @gosec_stderr = ''   # combined stderr on all runs (one for each configured subdir)
+      @gosec_stdout = ''   # ...      stdout ...
+      @gosec_json = {}     # combined json result on all runs
+      run_from_dirs = @config['run_from_dirs'].sort
+      @config.delete('run_from_dirs')
```
joshua-ostrom, line 19: Any need to restore this after the run_rom_dir blocks completes?  i.e. making this call impotent.

lib/salus/scanners/gosec.rb
ghbren: Great catch! Fixed.

lib/salus/scanners/gosec.rb
ghbren: Updated to restore the hash.

joshua-ostrom (approved):

PR #388, opened 2021-06-29
support for running bundle audit with local db

ghbren:
By default, BundleAudit pulls advisory info from [ruby-advisory-db](https://github.com/rubysec/ruby-advisory-db).
This PR adds support for scanning against a local advisory DB.  The config below will scan against the local DB in addition to ruby-advisory-db.
```yaml
scanner_configs:
  BundleAudit:
    local_db: $local_db_dir
```


joshua-ostrom (commented):
PR looks good. Do we foresee a case where we'd want to run only against the local advisory DB (and not dual run against the ruby-advisory-db too)?  If so, I could see a use case for a local-db-only:true config or similar.  But not need to be a blocker if it's a YAGNI.

joshua-ostrom (approved):

ghbren:
> PR looks good. Do we foresee a case where we'd want to run only against the local advisory DB (and not dual run against the ruby-advisory-db too)? If so, I could see a use case for a local-db-only:true config or similar. But not need to be a blocker if it's a YAGNI.

Good question. But I have not seen a case where we only want to run against the local DB.

joshua-ostrom (approved):

PR #396, opened 2021-07-13
Adding Salus enforced bool

saron-deb:
**What changed? Why?**
Adding Salus enforced bool under tool.driver.properties in SARIF in order to update Salus-bots display and cause less confusion surrounding active/enforced scanners

**Notes to reviewers**

**How has it been tested?**
Unit testing

**Helpful Links**

**General troubleshooting and FAQ**

**Change management (definitions)**
type=routine
risk=low
impact=sev5

**Automerge**
automerge=false


lib/sarif/base_sarif.rb
```diff
@@ -30,7 +30,10 @@ def build_tool(rules: [])
           "name" => @scan_report.scanner_name,
           "version" => @scan_report.version,
           "informationUri" => @uri,
-          "rules" => rules
+          "rules" => rules,
+          "properties" => {
+            "salusEnforced": @required || false
```
joshua-ostrom, line 35: IIRC the || false was to ensure we had proper support here for unsupported (scanners that don't yet have a sarif adapter present) scanners.  If that's the case can you include one spec for that case?  Beyond that looks great!

saron-deb: Yep that's the case, actually I just remembered, the unsupported scanner case is already being tested in base_sarif_spec.rb around lines 12-20. I'd be happy to add another spec if this is insufficient!

joshua-ostrom (dismissed):

joshua-ostrom (dismissed):

joshua-ostrom (approved):

PR #403, opened 2021-07-22
Enable Timeouts for Individual Scanners

cbged2:
This lets developers define a `scanner_timeout_s` key in scanner configurations. If a scanner is scanning for longer than `scanner_timeout_s` seconds, the scan times out and an error is reported accordingly. If no `scanner_timeout_s` is provided, the scan is allowed to run indefinitely. Also fixed some typos in `base_spec.rb`.

### DIFF
--- a/docs/scanners/base.md
+++ b/docs/scanners/base.md
@@ -37,17 +37,17 @@ scanner_configs:
 
 #### Maximum Lifespan
 
-At times, scanners may perform scans for unacceptable lengths of time. To limit this behavior, you can define `scanner-timeout-ms` with the number of milliseconds you wish the scan to last before it times out.  
+At times, scanners may perform scans for unacceptable lengths of time. To limit this behavior, you can define `scanner-timeout-s` with the number of seconds you wish the scan to last before it times out.  
 
 Example with `YarnAudit` configuration:
 
 ```yaml
 scanner_configs:
   YarnAudit:
-    scanner-timeout-ms: 60000
+    scanner-timeout-s: 60
 ```
 
-This will limit YarnAudit scans to 1 minute (60,000 milliseconds) in execution time.
+This will limit YarnAudit scans to 1 minute (60 seconds) in execution time.
 
 ## Reading/setting up custom configs for your scanner
 
--- a/docs/scanners/base.md
+++ b/docs/scanners/base.md
@@ -37,14 +37,14 @@ scanner_configs:
 
 #### Maximum Lifespan
 
-At times, scanners may perform scans for unacceptable lengths of time. To limit this behavior, you can define `max-lifespan` with the number of milliseconds you wish the scan to last before it times out.  
+At times, scanners may perform scans for unacceptable lengths of time. To limit this behavior, you can define `scanner-timeout-ms` with the number of milliseconds you wish the scan to last before it times out.  
 
 Example with `YarnAudit` configuration:
 
 ```yaml
 scanner_configs:
   YarnAudit:
-    max-lifespan: 60000
+    scanner-timeout-ms: 60000
 ```
 
 This will limit YarnAudit scans to 1 minute (60,000 milliseconds) in execution time.
--- a/docs/scanners/base.md
+++ b/docs/scanners/base.md
@@ -37,14 +37,14 @@ scanner_configs:
 
 #### Maximum Lifespan
 
-At times, scanners may perform scans for unacceptable lengths of time. To limit this behavior, you can define `scanner-timeout-s` with the number of seconds you wish the scan to last before it times out.  
+At times, scanners may perform scans for unacceptable lengths of time. To limit this behavior, you can define `scanner_timeout_s` with the number of seconds you wish the scan to last before it times out.  
 
 Example with `YarnAudit` configuration:
 
 ```yaml
 scanner_configs:
   YarnAudit:
-    scanner-timeout-s: 60
+    scanner_timeout_s: 60
 ```
 
 This will limit YarnAudit scans to 1 minute (60 seconds) in execution time.
--- a/docs/scanners/base.md
+++ b/docs/scanners/base.md
@@ -35,6 +35,20 @@ scanner_configs:
       Slack channel: #security
 ```
 
+#### Maximum Lifespan
+
+At times, scanners may perform scans for unacceptable lengths of time. To limit this behavior, you can define `max-lifespan` with the number of milliseconds you wish the scan to last before it times out.  
+
+Example with `YarnAudit` configuration:
+
+```yaml
+scanner_configs:
+  YarnAudit:
+    max-lifespan: 60000
+```
+
+This will limit YarnAudit scans to 1 minute (60,000 milliseconds) in execution time.
+
 ## Reading/setting up custom configs for your scanner
 
 For setting up custom configurations for your scanner, you can optionally use the helper method ```build_options```.
--- a/lib/salus/config.rb
+++ b/lib/salus/config.rb
@@ -38,7 +38,7 @@ module Salus
 
     DEFAULT_SCANNER_CONFIG = {
       'pass_on_raise' => false, # strong default - if a scanner raises, it counts as failure.
-      'scanner-timeout-s' => 0 # Scanner can run indefinitely if a max lifespan isn't provided
+      'scanner_timeout_s' => 0 # Scanner can run indefinitely if a max lifespan isn't provided
     }.freeze
 
     LOCAL_FILE_SCHEME_REGEX = /\Afile\z/.freeze # like file://foobar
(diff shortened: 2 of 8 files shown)


docs/scanners/base.md
```diff
@@ -35,6 +35,20 @@ scanner_configs:
...
+
+Example with `YarnAudit` configuration:
+
+```yaml
+scanner_configs:
+  YarnAudit:
+    scanner-timeout-ms: 60000
+```
+
+This will limit YarnAudit scans to 1 minute (60,000 milliseconds) in execution time.
```
ghbren, line 50: Could you change miliseconds to seconds?
We pretty much just want to ensure the scanner doesn't run over x minutes or something, so miliseconds is too precise and not really needed.

lib/salus/scanners/base.rb
```diff
@@ -76,11 +79,24 @@ def run!(salus_report:, required:, pass_on_raise:, reraise:)
...
+        end
 
         if @report.errors.any?
           pass_on_raise ? @report.pass : @report.fail
         end
+      rescue Timeout::Error => e
+        timeout_error_data = {
+          message: "Scanner #{name} timed out during execution",
+          error_class: ScannerTimeoutError,
+          backtrace: e.backtrace.take(5)
```
ghbren, line 93: why not just keep the whole backtrace?

lib/salus/scanners/base.rb
```diff
@@ -76,11 +79,24 @@ def run!(salus_report:, required:, pass_on_raise:, reraise:)
...
+        }
+        @report.error(timeout_error_data)
+        salus_report.error(timeout_error_data)
+
+        # Propagate this error if desired
+        raise ScannerTimeoutError, timeout_error_data.message if reraise
```
ghbren, line 99: If you raise an Exception, the salus will fail.
We don't want salus to fail if it times out.  We just want to log the timeout error.  Please also log with `bugsnag_notify`

lib/salus/config.rb
```diff
@@ -35,7 +35,8 @@ class Config
     ).freeze
 
     DEFAULT_SCANNER_CONFIG = {
-      'pass_on_raise' => false # strong default - if a scanner raises, it counts as failure.
+      'pass_on_raise' => false, # strong default - if a scanner raises, it counts as failure.
+      'scanner-timeout-ms' => 0 # Scanner can run indefinitely if a max lifespan isn't provided
```
ghbren, line 39: We don't really need the timeout here.
If timeout is not found in config, we'll just assume there is no timeout and run everything without the timeout stuff.

lib/salus/scanners/base.rb
```diff
@@ -76,11 +79,24 @@ def run!(salus_report:, required:, pass_on_raise:, reraise:)
       salus_report.add_scan_report(@report, required: required)
 
       begin
-        @report.record { run }
+        @report.record do
+          Timeout.timeout(max_lifespan / 1000.to_f) { run }
```
ghbren, line 83: best to not `run` in a Timeout block if no timeout is specified.

spec/lib/salus/processor_spec.rb
```diff
@@ -154,6 +154,7 @@ def remove_key(json_string, key = 'running_time')
           headers: { 'Content-Type' => 'application/json' },
           times: 1
         ) do |req|
+          puts req.body
```
ghbren, line 157: why `puts req.body` here?

lib/salus/scanners/base.rb
```diff
@@ -76,11 +79,24 @@ def run!(salus_report:, required:, pass_on_raise:, reraise:)
...
         end
+      rescue Timeout::Error => e
+        timeout_error_data = {
+          message: "Scanner #{name} timed out during execution",
+          error_class: ScannerTimeoutError,
+          backtrace: e.backtrace.take(5)
+        }
+        @report.error(timeout_error_data)
+        salus_report.error(timeout_error_data)
+
```
ghbren, line 97: I'm not sure if `error` is a callable method from `@report`/`salus_report`?

cbged2: Looks to be defined [here in report.rb](https://github.com/coinbase/salus/blob/master/lib/salus/report.rb#L72).

lib/salus/config.rb
cbged2: This does what you've mentioned. A timeout of 0 means that a scanner can run indefinitely (based on [the timeout docs](https://ruby-doc.org/stdlib-2.5.1/libdoc/timeout/rdoc/Timeout.html)).

lib/salus/scanners/base.rb
cbged2: Since the timeout [is defined as a default config value](https://github.com/coinbase/salus/pull/403/files/8f2aecf4fa15e4796d8328c4bc7bbf30ec93a386#diff-039372a0d92ae738bcced2074de46eb7f947085cd61a675a2e7aceb4b41cfb7e), a timeout is always specified.

cbged2:
I've addressed most of these changes in [this commit](https://github.com/coinbase/salus/commit/19d9e5ca369c644abf71a93de1c31f07096bd69d). For the others, I've responded to the review.

lib/salus/scanners/base.rb
```diff
@@ -76,11 +79,26 @@ def run!(salus_report:, required:, pass_on_raise:, reraise:)
...
 
         if @report.errors.any?
           pass_on_raise ? @report.pass : @report.fail
         end
+      rescue Timeout::Error
+        error_message = "Scanner #{name} timed out during execution"
+        timeout_error_data = {
+          message: error_message,
+          error_class: ScannerTimeoutError,
+          backtrace: []
```
ghbren, line 94: I think we should delete `backtrace: []`, because it's always a hardcoded empty array.

lib/salus/scanners/base.rb
```diff
@@ -76,11 +79,26 @@ def run!(salus_report:, required:, pass_on_raise:, reraise:)
...
-        @report.record { run }
+        @report.record do
+          Timeout.timeout(scanner_timeout) { run }
+        end
 
         if @report.errors.any?
           pass_on_raise ? @report.pass : @report.fail
         end
+      rescue Timeout::Error
+        error_message = "Scanner #{name} timed out during execution"
```
ghbren, line 90: could you append something like `after x seconds` to the message?

lib/salus/scanners/base.rb
```diff
@@ -461,5 +479,16 @@ def build_options(prefix:, suffix:, separator:, args:, join_by: ',')
         options + option
       end
     end
+
+    def scanner_timeout
+      scanner_timeout_config_param = @config['scanner_timeout_s']
+      # If a developer mistakenly defines this parameter
+      # as a non-integer value, let it be known
+      unless scanner_timeout_config_param.is_a? Integer
+        raise ConfigFormatError, "'scanner_timeout_s' parameter should be an integer"
```
ghbren, line 488: Could you `bugsnag_notify` the message too?

lib/salus/scanners/base.rb
```diff
@@ -461,5 +479,16 @@ def build_options(prefix:, suffix:, separator:, args:, join_by: ',')
         options + option
       end
     end
+
+    def scanner_timeout
+      scanner_timeout_config_param = @config['scanner_timeout_s']
+      # If a developer mistakenly defines this parameter
+      # as a non-integer value, let it be known
+      unless scanner_timeout_config_param.is_a? Integer
```
ghbren, line 487: What you have here is fine for now. 
But the `is_a? Integer` will also pass if a negative integer is used, and the code may behave strangely.

lib/salus/scanners/base.rb
```diff
@@ -76,11 +79,26 @@ def run!(salus_report:, required:, pass_on_raise:, reraise:)
...
+          message: error_message,
+          error_class: ScannerTimeoutError,
+          backtrace: []
+        }
+        @report.error(timeout_error_data)
+        salus_report.error(timeout_error_data)
+        bugsnag_notify(error_message)
+
+        # Propagate this error if desired
+        raise ScannerTimeoutError, timeout_error_data.message if reraise && !pass_on_raise
```
ghbren, line 101: Please delete ` && !pass_on_raise`.

Also, inside `rescue Timeout::Error` add at the breginning:
```ruby
        if @report.errors.any?
          pass_on_raise ? @report.pass : @report.fail
        end
        pass_on_raise ? @report.pass : @report.fail
```

In case you don't understand these changes, let me know.

ghbren: This would be a good spec to add:
Stub the `run` method of a scanner to let it `sleep 15`.  Then set timeout to `5` and check if salus reported timeout.

In case you don't know how to stub, please see https://www.tutorialspoint.com/rspec/rspec_stubs.htm

lib/salus/scanners/base.rb
ghbren: May change is to something like this?
`x.is_a? Integer && x >= 0`

ghbren (approved):

PR #405, opened 2021-07-25
ignore bad configs

ghbren:
Ignore configs that cannot be parsed as hashes and log.

### DIFF
--- a/lib/salus/config.rb
+++ b/lib/salus/config.rb
@@ -50,9 +50,11 @@ module Salus
       # defined in the earlier files in the array (DEFAULT_CONFIG is the first such file/object).
       final_config = DEFAULT_CONFIG.dup
       configuration_files.each do |file|
-        filtered_data = filter_ignored_ids(YAML.safe_load(file), ignore_ids)
-        if !filtered_data.is_a?(Hash)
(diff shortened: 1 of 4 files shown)


lib/salus/config.rb
```diff
@@ -48,7 +50,14 @@ def initialize(configuration_files = [], ignore_ids = [])
...
       final_config = DEFAULT_CONFIG.dup
       configuration_files.each do |file|
-        filtered_data = filter_ignored_ids(YAML.safe_load(file), ignore_ids)
+        file_content = YAML.safe_load(file)
+        if file_content.is_a?(Hash)
+          filtered_data = filter_ignored_ids(file_content, ignore_ids)
+        else
+          msg = "file_content #{file_content.inspect} is not a hash"
+          bugsnag_notify(msg)
+          filtered_data = {}
```
joshua-ostrom, line 59: We will cover a few additional bases if you move filtered_data = {} to line before the YAML_safe_load.  That will help if the safe_load runs into any classes not whitelisted (Disallowed classes).  Probably not a big deal not but would future proof the code here.

Rest looks good!

joshua-ostrom (approved):

PR #406, opened 2021-07-26
Add base CycloneDX structure and CycloneDX report files

jeffrey778zhan:
Initial work needed to set up base CycloneDX structure in which concrete scanners will then populate. Happy to create a CycloneDX feature branch where we can branch and merge CycloneDX changes into if we feel that's better.

### DIFF
--- a/lib/cyclonedx/cyclonedx_report.rb
+++ b/lib/cyclonedx/cyclonedx_report.rb
@@ -1,0 +1,55 @@
+require 'securerandom'
+
+Dir.entries(File.expand_path('./', __dir__)).sort.each do |filename|
+  next unless /_cyclonedx.rb\z/.match?(filename) && !filename.eql?('base_cyclonedx.rb')
+
+  require_relative filename
+end
+
+module Cyclonedx
+  class CyclonedxReport
+    def initialize(scan_reports, config = {})
+      @scan_reports = scan_reports
+      @config = config
+    end
+
+    CYCLONEDX_SPEC_VERSION = "1.2.0".freeze
+    CYCLONEDX_VERSION = "1".freeze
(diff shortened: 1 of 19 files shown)


lib/cyclonedx/report_go_dep_cyclonedx.rb
```diff
@@ -0,0 +1,7 @@
+module Cyclonedx
+  class ReportGoDepCyclonedx < BaseCyclonedx
```
ghbren, line 2: Perhaps it's better to rename the class to `ReportGoDep`?  
Otherwise, the class name is really `Cyclonedx:: ReportGoDepCyclonedx`, and the `Cyclonedx` at the end seems redudant.

lib/cyclonedx/cyclonedx_report.rb
```diff
@@ -0,0 +1,55 @@
+require 'securerandom'
+
+Dir.entries(File.expand_path('./', __dir__)).sort.each do |filename|
+  next unless /_cyclonedx.rb\z/.match?(filename) && !filename.eql?('base_cyclonedx.rb')
```
ghbren, line 4: where will you require `base_cyclonedx.rb`?  Probably need to ensure that gets required before the other files.

jeffrey778zhan: Good catch, added

lib/cyclonedx/report_go_dep_cyclonedx.rb
jeffrey778zhan: Yep I had it like so for consistency with Sarif `Sarif::BanditSarif` but agree that it's redundant. Happy to rename.

ghbren (approved):

ghbren (approved):

PR #408, opened 2021-07-28
Update pattern_search.rb

anantadityaCB:
Fixing the Salus Pattern Search Error where the Pattern search parser is not able to process files with larger line than 256KB

Bug | https://jira.coinbase-corp.com/browse/SECT-760

Added in a new variable with default size of 256K which can be changed for applications where the parsing is not working over to 512K and then re-processed.


anantadityaCB:
Removing whitespace

PR #409, opened 2021-07-28
Update pattern_search.rb

anantadityaCB:
Added in support to specify a blocksize attribute that will allow projects to increase the default size of Salus Pattern search scanner. Default is now set to 256KB

In order to increase we used 512KB to see if that allows the scanner to complete the processing or not. It worked with 512KB.


anantadityaCB:
Adding in the missing integration json to the commit

PR #410, opened 2021-07-28
CycloneDX Ruby Support

jeffrey778zhan:
Add Ruby Dependency Component builder with unit tests. Generate CycloneDX components follow spec version 1.3, specifically making use of the new properties type.

### DIFF
--- a/lib/cyclonedx/base.rb
+++ b/lib/cyclonedx/base.rb
@@ -1,6 +1,7 @@
 module Cyclonedx
   class Base
     DEFAULT_COMPONENT_TYPE = "application".freeze
+    DEFAULT_DEP_COMPONENT_TYPE = "library".freeze
 
     def initialize(scan_report, config = {})
       @scan_report = scan_report
@@ -29,6 +30,27 @@ module Cyclonedx
         components << parse_dependency(dependency)
       end
       components
+    end
+
+    def parse_dependency(dependency)
+      {
+        "bom-ref": package_url(dependency),
+        "type": DEFAULT_DEP_COMPONENT_TYPE,
+        "group": "", # TODO: add group or domain name of the publisher
+        "name": dependency[:name],
+        "version": dependency[:version],
+        "purl": package_url(dependency),
+        "properties": [
+          {
+            "key": "source",
+            "value": dependency[:source]
+          },
+          {
+            "key": "dependency_file",
+            "value": dependency[:dependency_file]
+          }
+        ]
+      }
     end
   end
 end
--- a/lib/cyclonedx/base.rb
+++ b/lib/cyclonedx/base.rb
@@ -38,7 +38,7 @@ module Cyclonedx
         "type": DEFAULT_DEP_COMPONENT_TYPE,
         "group": "", # TODO: add group or domain name of the publisher
         "name": dependency[:name],
-        "version": dependency[:version],
+        "version": version_string(dependency),
         "purl": package_url(dependency),
         "properties": [
           {
--- a/lib/cyclonedx/base.rb
+++ b/lib/cyclonedx/base.rb
@@ -24,18 +24,9 @@ module Cyclonedx
     # Returns the 'components' object for a supported/unsupported scanner's report
     def build_components_object
       components = []
-      @scan_report.info[:dependencies].each do |dependency|
-        component = {
-          "bom-ref": "",
-          "type": DEFAULT_COMPONENT_TYPE,
-          "group": "",
-          "name": dependency[:name],
-          "version": "",
-          "purl": ""
-        }
-
-        # TODO: Add specific component parsing for individual scanners
-        components << component
+      info = @scan_report.to_h.fetch(:info)
+      info[:dependencies].each do |dependency|
+        components << parse_dependency(dependency)
       end
       components
     end
--- a/lib/cyclonedx/report.rb
+++ b/lib/cyclonedx/report.rb
@@ -14,7 +14,7 @@ module Cyclonedx
       @config = config
     end
 
-    CYCLONEDX_SPEC_VERSION = "1.3.0".freeze
+    CYCLONEDX_SPEC_VERSION = "1.3".freeze
     CYCLONEDX_VERSION = "1".freeze
     CYCLONEDX_FORMAT = "CycloneDX".freeze
 
--- a/lib/cyclonedx/report.rb
+++ b/lib/cyclonedx/report.rb
@@ -14,7 +14,7 @@ module Cyclonedx
       @config = config
     end
 
-    CYCLONEDX_SPEC_VERSION = "1.2.0".freeze
+    CYCLONEDX_SPEC_VERSION = "1.3.0".freeze
     CYCLONEDX_VERSION = "1".freeze
     CYCLONEDX_FORMAT = "CycloneDX".freeze
 
--- a/lib/cyclonedx/report_ruby_gems_cyclonedx.rb
+++ b/lib/cyclonedx/report_ruby_gems_cyclonedx.rb
@@ -1,7 +1,30 @@
 module Cyclonedx
   class ReportRubyGems < Base
+    DEFAULT_COMPONENT_TYPE = "library".freeze
+
     def initialize(scan_report)
       super(scan_report)
+    end
+
+    def parse_dependency(dependency)
+      {
+        "bom-ref": "",
+        "type": DEFAULT_COMPONENT_TYPE,
+        "group": "", # TODO: add group or domain name of the publisher
+        "name": dependency[:name],
+        "version": dependency[:version],
+        "purl": "pkg:#{dependency[:type]}/#{dependency[:name]}@#{dependency[:version]}",
+        "properties": [
+          {
+            "key": "source",
+            "value": dependency[:source]
+          },
+          {
+            "key": "dependency_file",
+            "value": dependency[:dependency_file]
+          }
+        ]
+      }
     end
   end
 end
--- a/lib/cyclonedx/report_ruby_gems_cyclonedx.rb
+++ b/lib/cyclonedx/report_ruby_gems_cyclonedx.rb
@@ -5,14 +5,15 @@ module Cyclonedx
     end
 
     def package_url(dependency)
-      "pkg:#{dependency[:type]}/#{dependency[:name]}#{version_string(dependency)}"
+      "pkg:#{dependency[:type]}/#{dependency[:name]}#{version_string(dependency, true)}"
     end
 
     # Return version string to be used in purl
-    def version_string(dependency)
+    def version_string(dependency, is_purl_version = false)
+      prefix = is_purl_version ? "@" : ""
       if dependency[:dependency_file] == 'Gemfile.lock'
         # Return empty string if concrete dependency version specified in Gemfile.lock
-        "@#{dependency[:version]}"
+        "#{prefix}#{dependency[:version]}"
       else
         ""
       end
--- a/lib/cyclonedx/report_ruby_gems_cyclonedx.rb
+++ b/lib/cyclonedx/report_ruby_gems_cyclonedx.rb
@@ -1,30 +1,11 @@
 module Cyclonedx
   class ReportRubyGems < Base
-    DEFAULT_COMPONENT_TYPE = "library".freeze
-
     def initialize(scan_report)
       super(scan_report)
     end
 
-    def parse_dependency(dependency)
-      {
-        "bom-ref": "",
-        "type": DEFAULT_COMPONENT_TYPE,
-        "group": "", # TODO: add group or domain name of the publisher
-        "name": dependency[:name],
-        "version": dependency[:version],
-        "purl": "pkg:#{dependency[:type]}/#{dependency[:name]}#{version_string(dependency)}",
-        "properties": [
-          {
-            "key": "source",
-            "value": dependency[:source]
-          },
-          {
-            "key": "dependency_file",
-            "value": dependency[:dependency_file]
-          }
-        ]
-      }
+    def package_url(dependency)
+      "pkg:#{dependency[:type]}/#{dependency[:name]}#{version_string(dependency)}"
     end
 
     # Return version string to be used in purl
--- a/lib/cyclonedx/report_ruby_gems_cyclonedx.rb
+++ b/lib/cyclonedx/report_ruby_gems_cyclonedx.rb
@@ -13,7 +13,7 @@ module Cyclonedx
         "group": "", # TODO: add group or domain name of the publisher
         "name": dependency[:name],
         "version": dependency[:version],
-        "purl": "pkg:#{dependency[:type]}/#{dependency[:name]}@#{dependency[:version]}",
+        "purl": "pkg:#{dependency[:type]}/#{dependency[:name]}#{version_string(dependency)}",
         "properties": [
           {
             "key": "source",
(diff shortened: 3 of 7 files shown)


coderpatros:
Hi @jeffrey778zhan, noticed CycloneDX support was being added. Happy to provide some feedback if you like.

jeffrey778zhan:
> Hi @jeffrey778zhan, noticed CycloneDX support was being added. Happy to provide some feedback if you like.

Hey @coderpatros, absolutely! Any feedback is welcome.

lib/cyclonedx/report_ruby_gems_cyclonedx.rb
```diff
@@ -3,5 +3,19 @@ class ReportRubyGems < Base
     def initialize(scan_report)
       super(scan_report)
     end
+
+    def package_url(dependency)
```
jeffrey778zhan, line 7: For Ruby we can use the `dependency[:type]` but for others (i.e node) we'll need to change the type to match with the specs https://github.com/package-url/purl-spec

lib/cyclonedx/base.rb
```diff
@@ -24,20 +25,32 @@ def build_metadata
...
       components
     end
+
+    def parse_dependency(dependency)
+      {
+        "bom-ref": package_url(dependency),
+        "type": DEFAULT_DEP_COMPONENT_TYPE,
+        "group": "", # TODO: add group or domain name of the publisher
+        "name": dependency[:name],
+        "version": dependency[:version],
```
jeffrey778zhan: Version is a required field according to the [specs](https://github.com/CycloneDX/specification/blob/master/schema/bom-1.3.schema.json) will check against schema if empty string is valid.

nishils:
@jeffrey778zhan thanks for starting this up! 

Could you take a first stab at writing the docs for this scanner (and maybe some edits to the readme) before we merge and while we review the PR?

jeffrey778zhan:
> @jeffrey778zhan thanks for starting this up!
> 
> Could you take a first stab at writing the docs for this scanner (and maybe some edits to the readme) before we merge and while we review the PR?

Sure! I can add in the relevant information into salus_reports.md

lib/cyclonedx/report.rb
```diff
@@ -14,7 +14,7 @@ def initialize(scan_reports, config = {})
       @config = config
     end
 
-    CYCLONEDX_SPEC_VERSION = "1.2.0".freeze
+    CYCLONEDX_SPEC_VERSION = "1.3.0".freeze
```
coderpatros, line 17: The spec version should be `1.3`

lib/cyclonedx/base.rb
```diff
@@ -24,20 +25,32 @@ def build_metadata
...
+        "bom-ref": package_url(dependency),
+        "type": DEFAULT_DEP_COMPONENT_TYPE,
+        "group": "", # TODO: add group or domain name of the publisher
+        "name": dependency[:name],
+        "version": version_string(dependency),
+        "purl": package_url(dependency),
+        "properties": [
+          {
+            "key": "source",
+            "value": dependency[:source]
```
coderpatros, line 46: Does `dependency[:source]` always refer to the source code repository? If it does, a `vcs` type [external reference](https://cyclonedx.org/use-cases/#external-references) would be better than using a property.

lib/cyclonedx/base.rb
```diff
@@ -24,20 +25,32 @@ def build_metadata
...
+          },
+          {
+            "key": "dependency_file",
+            "value": dependency[:dependency_file]
```
coderpatros, line 50: There is an option to have a `build-meta` external reference if you wanted to provide a way to retrieve the gemfile that was used.

lib/cyclonedx/base.rb
coderpatros: Ah, I just noticed further down that this might not always be the case. If the source does refer to a repo it would be good to include it as an external reference.

lib/cyclonedx/report_ruby_gems_cyclonedx.rb
```diff
@@ -3,5 +3,20 @@ class ReportRubyGems < Base
...
+    # Return version string to be used in purl
+    def version_string(dependency, is_purl_version = false)
+      prefix = is_purl_version ? "@" : ""
+      if dependency[:dependency_file] == 'Gemfile.lock'
+        # Return empty string if concrete dependency version specified in Gemfile.lock
+        "#{prefix}#{dependency[:version]}"
+      else
+        ""
+      end
+    end
```
coderpatros, line 20: It would be really good to use the version of the resolved gem that is actually being used over what is in the gemfile. That's part of the purpose of CycloneDX.

coderpatros: Ah, just re-read the docs on how Salus works. I don't know how you can accurately determine package versions outside of a build process.

lib/cyclonedx/report.rb
jeffrey778zhan: Thanks 👍

lib/cyclonedx/report_ruby_gems_cyclonedx.rb
jeffrey778zhan: Yep, we're currently using the Gemfile.lock to determine the exact version but in the case of just a Gemfile that's provided we don't include the version.

lib/cyclonedx/base.rb
jeffrey778zhan: Yeah, I think in this case the `dependency[:source]` is different from the source code repository, but will definitely keep this in mind if/when we do want to include that information

lib/cyclonedx/base.rb
jeffrey778zhan: Thanks for the insight! This property is currently being used to specify which file the dependencies were parsed from (i.e Gemfile vs Gemfile.lock). Will keep the build-meta external reference in mind if we want to provide a url to the dependency file used.

lib/cyclonedx/base.rb
jeffrey778zhan: Confirmed empty version is valid

ghbren:
Does CycloneDX provide a function for you to validate the result against a schema?

spec/lib/cyclonedx/report_ruby_gems_cyclonedx_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+
+      ruby_cyclonedx = Cyclonedx::ReportRubyGems.new(scanner.report)
+      expect(ruby_cyclonedx.build_components_object).to match_array(
+        [
+          {
+            "bom-ref": "pkg:gem/kibana_url",
+            "type": "library",
+            "group": "",
+            "name": "kibana_url",
+            "version": "",
```
ghbren, line 19: The Gemfile says `gem 'kibana_url', '~> 1.0'`, so should. you add `~> 1.0` to `version`?

spec/lib/cyclonedx/report_ruby_gems_cyclonedx_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+            "bom-ref": "pkg:gem/master_lock",
+            "type": "library",
+            "group": "",
+            "name": "master_lock",
+            "version": "",
+            "purl": "pkg:gem/master_lock",
+            "properties": [
+              {
+                "key": "source",
+                "value": "git@github.com:coinbase/master_lock.git"
```
ghbren, line 60: The Gemfile says `gem 'master_lock', git: 'git@github.com:coinbase/master_lock.git'`.
Does `git@github.com:coinbase/master_lock.git` specify the version?  If so, should you add it to `version`?

spec/lib/cyclonedx/report_ruby_gems_cyclonedx_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+              "value": "git@github.com:coinbase/master_lock.git"
+            },
+            {
+              "key": "dependency_file",
+              "value": "Gemfile.lock"
+            }
+          ]
+        }
+      ]
+      expect(ruby_cyclonedx.build_components_object).to include(*expected)
```
ghbren, line 152: This is part of the Gemfile.lock
`
  specs:
    actioncable (5.1.2)
      actionpack (= 5.1.2)
      nio4r (~> 2.0)
  ...
`
Are nested deps like `actionpack`, `nio4r` included in the output too?  If so, could you add one of these?

spec/lib/cyclonedx/report_ruby_gems_cyclonedx_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+            "bom-ref": "pkg:gem/kibana_url",
+            "type": "library",
+            "group": "",
+            "name": "kibana_url",
+            "version": "",
+            "purl": "pkg:gem/kibana_url",
+            "properties": [
+              {
+                "key": "source",
+                "value": "https://rubygems.org/"
```
ghbren, line 24: Seems like all of your test cases use `https://rubygems.org` as the only gem source.
Some Gemfile.locks have multiple sources.  Have you tested with these?  
This would be good as an additional test case.

ghbren (commented):
.

coderpatros:
> Does CycloneDX provide a function for you to validate the result against a schema?

You can use the CycloneDX [CLI tool](https://github.com/CycloneDX/cyclonedx-cli/) or [web tool](https://cyclonedx.github.io/cyclonedx-web-tool/) for validation.

Alternatively you can grab a copy of the schemas from the [specification repo](https://github.com/CycloneDX/specification/tree/master/schema) if you want to perform your own validation.

ghbren:
> Alternatively you can grab a copy of the schemas from the [specification repo](https://github.com/CycloneDX/specification/tree/master/schema) if you want to perform your own validation.

Jeffrey, we included a copy of the sarif schema to validate against our sarif results.  Can you look into whether it's possible to do something similar for CycloneDX?

jeffrey778zhan:
> Does CycloneDX provide a function for you to validate the result against a schema?

Yes, https://github.com/coinbase/salus/pull/415 

> Does CycloneDX provide a function for you to validate the result against a schema?

Yep, PR to integrate that here https://github.com/coinbase/salus/pull/415

lib/cyclonedx/report_ruby_gems_cyclonedx.rb
```diff
@@ -3,5 +3,20 @@ class ReportRubyGems < Base
...
       super(scan_report)
     end
+
+    def package_url(dependency)
+      "pkg:#{dependency[:type]}/#{dependency[:name]}#{version_string(dependency, true)}"
+    end
+
+    # Return version string to be used in purl
+    def version_string(dependency, is_purl_version = false)
+      prefix = is_purl_version ? "@" : ""
```
joshua-ostrom, line 13: Minor suggestion.  A guard class up top is preferred to the conditional if else block / flows:

```
return "" unless dependency[:dependency_file] == 'Gemfile.lock'
prefix = is_purl_version ? "@" : ""
"#{prefix}#{dependency[:version]}"
```

spec/lib/cyclonedx/report_ruby_gems_cyclonedx_spec.rb
jeffrey778zhan: Right, ideally the actual version of the Gem should be specified in CycloneDX in semantic versioning format according to the [schema](https://github.com/CycloneDX/specification/blob/master/schema/bom-1.3.schema.json#L291). In the schema it simply requires the version to be a string so we could add a pessimistic version strings as well.

spec/lib/cyclonedx/report_ruby_gems_cyclonedx_spec.rb
jeffrey778zhan: I believe `git@github.com:coinbase/master_lock.git` specifies the version in its `gem.spec` file (i.e https://github.com/coinbase/master_lock/blob/master/master_lock.gemspec) but we don't have access to the exact version used from just the Gemfile but a constraint on the version (i.e `>= 0`) which we can include if desired.

spec/lib/cyclonedx/report_ruby_gems_cyclonedx_spec.rb
jeffrey778zhan: Yep they are included 👍 `actioncable` itself is a nested dependency of rails. Happy to add in some more examples to the spec.

spec/lib/cyclonedx/report_ruby_gems_cyclonedx_spec.rb
jeffrey778zhan: Good idea, will add!

jeffrey778zhan:
> Have you tested something like `docker build ...`, then `docker run ...` to generate the cycloneDX output file, to ensure no strange behavior?

Yep, if I'm understanding your question correctly currently this PR is scoped for only the changes needed to build CycloneDX reports from Ruby dependencies so there's still config changes needed to be made (i.e to add option to read the cyclonedx format option in the config) before we can generate the cycloneDX output file from docker. And at that point we'll be ready to test something like `docker build ...`, then `docker run ...`

lib/cyclonedx/report_ruby_gems_cyclonedx.rb
jeffrey778zhan: Nice, will make the change 👍

jeffrey778zhan:
Hi @coderpatros, for the `purl` [field](https://github.com/CycloneDX/specification/blob/master/schema/bom-1.3.schema.json#L335) is the version required to be absolute? For example, would something like`pkg:maven/com.acme/tomcat-catalina@~> 9.0` be a valid purl?

coderpatros:
> Hi @coderpatros, for the `purl` [field](https://github.com/CycloneDX/specification/blob/master/schema/bom-1.3.schema.json?rgh-link-date=2021-08-02T15%3A04%3A23Z#L335) is the version required to be absolute? For example, would something like`pkg:maven/com.acme/tomcat-catalina@~> 9.0` be a valid purl? Thanks!

I don't think it would be valid. Besides the version not being encoded, I think it should be a specific version only or omitted. @pombredanne @stevespringett is that statement correct for purls?

pombredanne:
@jeffrey778zhan you wrote:
> > Hi @coderpatros, for the `purl` [field](https://github.com/CycloneDX/specification/blob/master/schema/bom-1.3.schema.json?rgh-link-date=2021-08-02T15%3A04%3A23Z#L335) is the version required to be absolute? For example, would something like `pkg:maven/com.acme/tomcat-catalina@~> 9.0` be a valid purl? Thanks!
> 
> I don't think it would be valid. Besides the version not being encoded, I think it should be a specific version only or omitted. @pombredanne @stevespringett is that statement correct for purls?

Short answer: a version need to be a version, e.g. a resolved concrete version and not a range, constraints, wildcard or specifier of sorts. `~> 9.0` is a version constraint/range/specifier and this is not meant to be used as a version in a purl.

So:
- `pkg:maven/com.acme/tomcat-catalina@~> 9.0` is not valid
- `pkg:maven/com.acme/tomcat-catalina@9.0` would be valid
- `pkg:maven/com.acme/tomcat-catalina@%7E%3E%209.0` would be technically valid encoding-wise but ugly and impossible to resolve since `~> 9.0` is not a version, so this is also invalid.


Eventually we want to a spec how to encode version ranges likely as a purl qualifier and there is work underway.
- Key discussions are at https://github.com/package-url/purl-spec/issues/66 and  https://github.com/package-url/purl-spec/issues/84
- There is a  rejected proposal here: https://github.com/package-url/purl-spec/pull/93
- Some work is in progress with code experiments in https://github.com/nexB/vulnerablecode/issues/140 and https://github.com/nexB/univers

The general directions I favor is explained at https://github.com/nexB/vulnerablecode/issues/119#issuecomment-799365438 and https://github.com/package-url/purl-spec/issues/84#issuecomment-831286705

joshua-ostrom (approved):

PR #411, opened 2021-07-29
Add support for go module dependency tracking

saron-deb:
Updating Salus' reporting of Golang dependencies to support reporting dependencies through modules.


lib/salus/scanners/report_go_dep.rb
```diff
@@ -7,38 +7,66 @@
...
           InvalidScannerInvocationError,
-          'Cannot report on Go dependencies without a Gopkg.lock file.'
+          'Cannot report on Go dependencies without a Gopkg.lock or go.mod/go.sum file'
         )
       end
+    end
+
+    def record_dep_from_go_mod
+      Dir.chdir(@repository.path_to_repo.to_s) do
+        shell_return = run_shell("go list -json -m all")
```
jeffrey778zhan, line 24: Do we know if go list accounts for transitive dependencies? I found this article here which provides some details on that https://dave.cheney.net/2014/09/14/go-list-your-swiss-army-knife but if it does that could be an easy win.

spec/lib/salus/scanners/report_go_dep_spec.rb
```diff
@@ -48,6 +48,43 @@
     end
   end
 
+  describe '#record_dep_from_go_mod' do
```
jeffrey778zhan, line 51: 💯 I think it would be a great opportunity to add some of these test to confirm if `go list` can handle
- internal modules
- both go.mod and go.sum
- if it accounts for transitive dependencies

saron-deb: Good points! 

I think after testing extensively it seems that internal modules were a lot of trouble to deal with when using `go list` so instead I just parsed go.sum / go.mod manually in addition to the existing Gopkg.lock as were thinking about doing before.

lib/salus/scanners/report_go_dep.rb
```diff
@@ -7,38 +7,106 @@
...
         )
       end
+    end
+
+    def record_dep_from_go_sum
+      go_sum_path = "#{@repository.path_to_repo}/go.sum"
+      dep_list = []
+
+      File.open(go_sum_path).each("=\n") do |line|
+        dep_list.append({ "fullDependency" => line, "name" => get_name(line),
```
jeffrey778zhan, line 29: Nit: For readability can we do something along the lines of
```
dep_list.append(
 { 
   "fullDependency" => line, 
   "name" => get_name(line), 
   "version" => get_version(line) 
 }
)

File: lib/salus/scanners/report_go_dep.rb [hunk 4]
```diff
@@ -7,38 +7,106 @@
...
+      dep_list = []
+      parse_line = false
+
+      File.open(go_mod_path).each("\n") do |line|
+        if line.include? "require ("
+          parse_line = true
+        elsif line.include? ")"
+          parse_line = false
+        elsif parse_line
+          dep_list.append({ "fullDependency" => line, "name" => get_name(line),
```
jeffrey778zhan, line 55: Same here

lib/salus/scanners/report_go_dep.rb
```diff
@@ -7,38 +7,106 @@
...
+        )
+      end
+    end
+
+    def record_dep_from_go_mod
+      go_mod_path = "#{@repository.path_to_repo}/go.mod"
+      dep_list = []
+      parse_line = false
+
+      File.open(go_mod_path).each("\n") do |line|
```
jeffrey778zhan, line 49: This is likely an edge case but for some go.mod files with only one package such as https://sourcegraph.cbhq.net/github.cbhq.net/c3/go-shamir/-/blob/go.mod it doesn't follow the `require (`. We may want to be careful there.

jeffrey778zhan (commented):
Looks good! Added some comments/questions

lib/salus/scanners/report_go_dep.rb
saron-deb: Yep, changed 👍

lib/salus/scanners/report_go_dep.rb
saron-deb: Yep 👍

lib/salus/scanners/report_go_dep.rb
```diff
@@ -7,38 +7,126 @@
...
+          'Cannot report on Go dependencies without a Gopkg.lock, go.mod, or go.sum file'
         )
       end
+    end
+
+    def record_dep_from_go_sum
+      go_sum_path = "#{@repository.path_to_repo}/go.sum"
+      dep_list = []
+
+      File.open(go_sum_path).each("=\n") do |line|
```
joshua-ostrom, line 28: File.open will slurp the entire file into memory, File.foreach will enumerate over each line

lib/salus/scanners/report_go_dep.rb
```diff
@@ -7,38 +7,126 @@
...
+        dep_list.append(
```
joshua-ostrom, line 29: probably not a bad idea to skip if the line is empty

lib/salus/scanners/report_go_dep.rb
```diff
@@ -7,38 +7,126 @@
...
+          {
+            "fullDependency" => line,
+          "name" => get_name(line),
```
joshua-ostrom, line 32: if you made one function to parse the line info you could avoid splitting the string twice.

lib/salus/scanners/report_go_dep.rb
```diff
@@ -7,38 +7,126 @@
...
+          "version" => get_version(line)
+          }
+        )
+      end
+
+      dep_list.each do |dependency|
+        record_dep_package(
+          name: dependency['name'],
+          reference: "N/A for go.mod/go.sum dependencies",
```
joshua-ostrom, line 41: maybe include a comment as to why references are NA here

lib/salus/scanners/report_go_dep.rb
joshua-ostrom: File.foreach to avoid a slurp here too.  It's a minor nit as the go.sum likely won't grow unbounded but a little cleaner.

joshua-ostrom (commented):
Safe to assume only go.sum flows will catch / report transitive dependencies here?  (I didn't see a go list used - any background on where you landed here ?)

lib/salus/scanners/report_go_dep.rb
```diff
@@ -7,38 +7,126 @@
...
-          version_tag: dependency['version']
+          version_tag: dependency['version'],
+          dependency_file: "Gopkg.lock",
+          type: "go_dep_lock"
         )
       end
     end
 
-    def should_run?
-      @repository.dep_lock_present?
```
joshua-ostrom, line 29: Any reason not to expand this to include the other checks vs dropping it?

joshua-ostrom: https://github.com/coinbase/salus/blob/23cef116093ab6736aaea48be8e55ddea17f6c07/lib/salus/processor.rb#L77. This is called outside of the logic here as part of our interface the scanner should implement

saron-deb: Actually I think I deleted this and rewrote it with the other checks around line 121 in report_go_dep.rb :sweat_smile:

saron-deb:
> Safe to assume only go.sum flows will catch / report transitive dependencies here? (I didn't see a go list used - any background on where you landed here ?)

Yes that's right, go.sum flows will catch all transitive dependencies. 
For go.mod with transitive dependencies, it will record them as referenced below:

"Any indirect dependency which is not listed in the go.mod file of your direct dependency or if direct dependency doesn’t have a go.mod file , then that dependency will be added to the go.mod file with //indirect as the suffix" 

As for using go list, I decided against it because it seemed like it would be more trouble to integrate with internal dependencies. I found it was possible by changing the (git config/go env) as talked about previously, but integrating that with docker seemed like it may cause risks as it seemed like a personal github access token is needed.

Happy to go back and look into this more if needed 👍

lib/salus/scanners/report_go_dep.rb
saron-deb: Changed 👍

lib/salus/scanners/report_go_dep.rb
saron-deb: Changed 👍

lib/salus/scanners/report_go_dep.rb
saron-deb: Added!

lib/salus/scanners/report_go_dep.rb
saron-deb: Refactored for less splitting 👍

lib/salus/scanners/report_go_dep.rb
saron-deb: Added!

lib/salus/scanners/report_go_dep.rb
ghbren: What lines with just a tab or something?
I think you should use `String#strip` https://ruby-doc.org/core-3.0.2/String.html#method-i-strip

lib/salus/scanners/report_go_dep.rb
```diff
@@ -7,38 +7,131 @@
...
+
+    def record_dep_from_go_mod
+      go_mod_path = "#{@repository.path_to_repo}/go.mod"
+      dep_list = []
+      parse_line = false
+
+      File.foreach(go_mod_path).each("\n") do |line|
+        next if line.empty?
+
+        if line.include? "require ("
```
ghbren, line 61: It would be more precise to add a check to eliminate cases like `XYZrequire ("

Also, there could be spaces between `require` and `(`. So it would be best to use a regular expression for the check.

lib/salus/scanners/report_go_dep.rb
```diff
@@ -7,38 +7,131 @@
...
+          parse_line = true
+        # Edge case with only 1 dependency in go.mod
+        elsif line.include? "require "
```
ghbren, line 64: Is this syntax also valid in go dep files?
```require(github.com...```

File: lib/salus/scanners/report_go_dep.rb [hunk 13]
```diff
@@ -7,38 +7,131 @@
...
+
+          dep_list.append(
+            {
+              "fullDependency" => line_info[1] + line_info[2],
+              "name" => line_info[1],
+              "version" => line_info[2]
+            }
+          )
+        elsif line.include? ")"
+          parse_line = false
```
ghbren, line 75: Is this valid syntax too?
```github.com/stretchr/testify v1.7.0)```

File: lib/salus/scanners/report_go_dep.rb [hunk 14]
```diff
@@ -7,38 +7,131 @@
...
         )
       end
+    end
+
+    def record_dep_from_go_sum
+      go_sum_path = "#{@repository.path_to_repo}/go.sum"
+      dep_list = []
+
+      File.foreach(go_sum_path).each("=\n") do |line|
+        next if line.empty?
```
ghbren, line 29: What about lines with other types whitespaces, like as pre/suffixes?
I think you should use String#strip before checking if line is empty
https://ruby-doc.org/core-3.0.2/String.html#method-i-strip

Same for your other record_ methods.

lib/salus/scanners/report_go_dep.rb
ghbren, line 64: Well, this 
```ruby
elsif line.include? "require "
``` could match something like 
```
module github.com/x/y/z/my_project_called_require  # a comment
```

File: lib/salus/scanners/report_go_dep.rb (hunk 11 above)
Comment (ghbren) [reply lib/salus/scanners/report_go_dep.rb:61]: Moreover, what if `require (` is just part of a comment?

Comment (saron-deb) [reply lib/salus/scanners/report_go_dep.rb:61]: Good point, I'll strip each line first and switch this check out with a regex like `/^require\s{0,} \(/`.

I think that will cover cases like with whitespaces preceding or between require and (. Cases like `XYZrequire (` and comments should be covered too as they start with '#' or similar.

File: lib/salus/scanners/report_go_dep.rb (hunk 12 above)
Comment (saron-deb) [reply lib/salus/scanners/report_go_dep.rb:64]: Will change to a regex like `^require\s{0,} ` to avoid any matching other than the beginning of the line 👍

File: lib/salus/scanners/report_go_dep.rb (hunk 14 above)
Comment (saron-deb) [reply lib/salus/scanners/report_go_dep.rb:30]: Changed 👍

File: lib/salus/scanners/report_go_dep.rb (hunk 12 above)
Comment (saron-deb) [reply lib/salus/scanners/report_go_dep.rb]: I don't think so as per (https://golang.org/ref/mod#go-mod-file) not 100% sure on this though, can't find too much information on required syntax.

File: lib/salus/scanners/report_go_dep.rb (hunk 13 above)
Comment (saron-deb) [reply lib/salus/scanners/report_go_dep.rb:77]: I don't believe it is from (https://golang.org/ref/mod#go-mod-file) at least.

File: lib/salus/scanners/report_go_dep.rb (hunk 10 above)
Comment (joshua-ostrom) [reply lib/salus/scanners/report_go_dep.rb:29]: ah thanks!

File: lib/salus/scanners/report_go_dep.rb [hunk 15]
```diff
@@ -1,44 +1,105 @@
...
+require 'salus/plugin_manager'
+require 'salus/report'
 
 # Report the use of Go packages captured in a Gopkg.lock files.
 # https://github.com/golang/dep
 
 module Salus::Scanners
   class ReportGoDep < Base
     def run
-      unless should_run?
```
joshua-ostrom, line 10: any reason the "unless should_run?" is being dropped?

spec/lib/salus/scanners/report_go_dep_spec.rb
```diff
@@ -48,6 +48,72 @@
...
+            name: 'github.cbhq.net/c3/bls12-381',
+            reference: 'N/A for go.mod/go.sum dependencies',
+            version_tag: "v0.0.0-20210114210818-577bfdc5cb9c"
+          },
+          {
+            dependency_file: 'go.sum',
+            type: 'golang',
+            name: 'github.cbhq.net/c3/bls12-381',
+            reference: 'N/A for go.mod/go.sum dependencies',
+            version_tag: "v0.0.0-20210114210818-577bfdc5cb9c/go.mod"
```
joshua-ostrom, line 74: The /go.mod should probably be dropped from the version here.  I think we'll want to abstract the go.sum formats away from the output here.

lib/salus/scanners/report_go_dep.rb
```diff
@@ -1,44 +1,105 @@
...
-    def should_run?
-      @repository.dep_lock_present?
-    end
-
     def self.supported_languages
       ['go']
     end
 
-    private
+    def get_line_info(line)
```
joshua-ostrom, line 88: There are two versions of go.sum entries.  One with the /go.mod at the end of the version and one without.

In place of exploding any content on the space setting up an expected regex to match entries against might be a nice win:

```
# Sample
line = "github.com/go-ozzo/ozzo-validation v3.6.0+incompatible/go.mod h1:gsEKFIVnabGBt6mXmxK0MoFy+cZoTJY6mu5Ll3LVLBU="
go_sum_regex = /(?<namespace>(.*)(?!\/go\.mod))\/(?<name>[^\s]*)(\s)*(?<version>(.*))(\s)*h1:(?<checksum>(.*))/
if matches = line.match(go_sum_regex)
   puts "Namespace: #{matches[:namespace]}"
   puts "Name: #{matches[:name]}"
   puts "Version: #{matches[:version]}"
   puts "Checksum: #{matches[:checksum]}"
end
```
That would normalize the output so the version aligns between the 2 go.sum formats here.  It would also allow you to detect if you had unexpected formats present.

lib/salus/scanners/report_go_dep.rb
```diff
@@ -1,44 +1,105 @@
...
+      File.foreach(go_sum_path).each("=\n") do |line|
+        line = line.strip
+        next if line.empty?
+
+        line_info = get_line_info(line)
+
+        dep_list.append(
+          {
+            "fullDependency" => line,
+            "name" => line_info[0],
```
joshua-ostrom, line 39: See note for a cleaner way to reference name / version vs the positional reference here

lib/salus/scanners/report_go_dep.rb
```diff
@@ -1,44 +1,105 @@
...
+      # Typically go.mod files don't contain all of the transitive deps/info
+      # a go.sum file would. Instead of parsing go.mod files, warn that
+      # go.sum is missing and send an event
+      warning_string = "WARNING: No go.sum/Gopkg.lock found. "\
+      "Currently go.mod is unsupported for reporting Golang dependencies."
+      report_warn(:report_go_dep_missing_go_sum, warning_string)
+
+      data = "This repository contains no go.sum or Gopkg.lock file. Currently "\
+      "go.mod files are unsupported for reporting Golang dependencies"
+      Salus::PluginManager.send_event(:report_go_dep_scan, data)
```
joshua-ostrom, line 66: Might be a nice win to move the plugin hook here to within the report_warn. That'd expand the usability to all report warns.

lib/salus/scanners/report_go_dep.rb
saron-deb: I was just a little worried it might be redundant w/the conditionals below covering the same ground - added it back!

lib/salus/scanners/report_go_dep.rb
```diff
@@ -1,47 +1,124 @@
...
+          )
+        end
+      end
+      # Note references are hashes meant for packages in Gopkg.lock files
+      dep_list.each do |dependency|
+        record_dep_package(
+          namespace: dependency["namespace"],
+          name: dependency["name"],
+          reference: "N/A for go.mod/go.sum dependencies",
+          version_tag: dependency['version'].gsub(%r{/go.mod}, '').strip,
```
joshua-ostrom, line 56: Let's move the .gsub(%r{/go.mod}, '').strip up to line 44 👍

lib/salus/scanners/report_go_dep.rb
```diff
@@ -1,47 +1,124 @@
...
+      if @repository.go_sum_present?
+        record_dep_from_go_sum
+      elsif @repository.dep_lock_present?
+        record_dep_from_go_lock_package
+      elsif @repository.go_mod_present?
+        record_dep_from_go_mod
+      else
         raise(
           InvalidScannerInvocationError,
-          'Cannot report on Go dependencies without a Gopkg.lock file.'
```
joshua-ostrom, line 13: you can drop this final else block now that the unless should_run? is present

lib/salus/scanners/report_go_dep.rb
```diff
@@ -1,47 +1,119 @@
...
+      # Typically go.mod files don't contain all of the transitive deps/info
+      # a go.sum file would. Instead of parsing go.mod files, warn that
+      # go.sum is missing and send an event
+      warning_string = "WARNING: No go.sum/Gopkg.lock found. "\
+      "Currently go.mod is unsupported for reporting Golang dependencies."
+      report_warn(:report_go_dep_missing_go_sum, warning_string)
+
+      data = "This repository contains no go.sum or Gopkg.lock file. Currently "\
+      "go.mod files are unsupported for reporting Golang dependencies."
+      Salus::PluginManager.send_event(:report_go_dep_missing_go_sum, data)
```
joshua-ostrom, line 69: You can abstract the send event here to begin sending events for all report_warns.  That'll let you drop lines 67-69 here and provide a general report warn support for plugins.  Just be sure to add an entry for this event in https://github.com/coinbase/salus/blob/03279595de981a0aaad278f90f51d797e728e348/docs/architecture.md

`Salus::PluginManager.send_event(:report_warn, {type: type, message: message)
`
```
# Report a scanner warning such as a possible misconfiguration
    def report_warn(type, message)
      @report.warn(type, message)
      Salus::PluginManager.send_event(:report_warn, {type: type, message: message)
      if @salus_report&.builds
        scanner = @report.scanner_name
        message = "#{scanner} warning: #{type}, #{message}, build: #{@salus_report.builds}"
      end
      bugsnag_notify(message)
    end
```

joshua-ostrom (approved):

PR #412, opened 2021-07-29
Added support for defining blocksize for Pattern search

anantadityaCB:
Added in support to specify a blocksize attribute that will allow projects to increase the default size of Salus Pattern search scanner. Default is now set to 256KBIn order to increase we used 512KB to see if that allows the scanner to complete the processing or not. It worked with 512KB.

### DIFF
--- a/lib/salus/scanners/pattern_search.rb
+++ b/lib/salus/scanners/pattern_search.rb
@@ -47,7 +47,8 @@ module Salus::Scanners
           # if both are specified, they should be joined
           ex_paths = match['exclude_filepaths'] || @config['exclude_filepaths']
           exclude_filepath_pattern = filepath_pattern(ex_paths)
-
+          # Setting default block size in kilobytes to 256KB
+          match['blocksize'] ||= '256K'
           command_array = [
             "sift",
             "-n",
@@ -60,7 +61,9 @@ module Salus::Scanners
             ".",
             *(match_exclude_directory_flags || global_exclude_directory_flags),
             *(match_exclude_extension_flags || global_exclude_extension_flags),
-            *(match_include_extension_flags || global_include_extension_flags)
+            *(match_include_extension_flags || global_include_extension_flags),
+            "--blocksize",
+            match['blocksize']
           ].compact
 
           shell_return = run_shell(command_array)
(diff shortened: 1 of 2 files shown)


lib/salus/scanners/pattern_search.rb
```diff
@@ -60,7 +61,9 @@ def run
             ".",
             *(match_exclude_directory_flags || global_exclude_directory_flags),
             *(match_exclude_extension_flags || global_exclude_extension_flags),
-            *(match_include_extension_flags || global_include_extension_flags)
+            *(match_include_extension_flags || global_include_extension_flags),
+            "--blocksize",
```
joshua-ostrom, line 65: Be sure to document in docs/scanners/pattern_search.md as part of this PR

Also adding support for err-show-line-length and err-skip-line-length might be a nice win here.

```
['err-show-line-length', 'err-skip-line-length'].each do |flag|
	command_array << "--#{flag }"if match.key?(flag')
end

```
That would allow users of Salus to not only set their desired threshold via blocksize but allow them to disable the error should the linelength exceed blocksize.

docs/scanners/pattern_search.md
```diff
@@ -9,7 +9,7 @@ The scanner also allows the options below.  These options can be set globally an
 * `exclude_extension` and `include_extension` for excluding and including file extensions, respectively. While these options can be combined, exclusions take precedence when extensions conflict (are both included and excluded) in declarations.
 * `exclude_directory` for excluding directories whose name matches GLOB.  It appears that sift does not support `/`s in the directory name.
 * `exclude_filepaths` for excluding file paths. Note the file paths must be regular paths, not GLOB, and cannot include regular expressions.
-
+* `blocksize` for defining the amount of memory that will be used to limit the parser. By default this is set to 256 KB and can be increased based on the application needs
```
joshua-ostrom, line 12: Small copy suggestions:

Add "blocksize in bytes (with optional suffix K|M)"

"By default this is set to 256KB and can be increased" => "By default this is set to 256K and can be increased" to avoid confusion on the format of the default value.

```
* `blocksize` - (blocksize in bytes with optional suffix K|M) for defining the amount of memory that will be used to limit the parser.   By default this is set to 256K and can be increased based on the application needs

```

joshua-ostrom (approved):
Approved, some possible small copy tweaks but should be good to go.

PR #414, opened 2021-07-30
Enable Salus-wide Execution Timeout

cbged2:
This lets developers define a `salus_timeout_s` key in a Salus configuration. If Salus is scanning for longer than `salus_timeout_s` seconds, it times out and an error is reported accordingly. If no `salus_timeout_s` is provided, Salus can run indefinitely.


lib/salus/processor.rb
```diff
@@ -80,35 +82,46 @@ def fetch_config_file(source_uri, repo_path)
...
+          reraise_exceptions = ENV.key?('RUNNING_SALUS_TESTS')
+          Config::SCANNERS.each do |scanner_name, scanner_class|
+            config = @config.scanner_configs.fetch(scanner_name, {})
+
+            scanner = scanner_class.new(repository: repo, config: config)
+            unless @config.scanner_active?(scanner_name) && scanner.should_run?
+              Salus::PluginManager.send_event(:skip_scanner, scanner_name)
+              next
+            end
+            scanners_ran << scanner
```
ghbren, line 100: You should move `scanners_ran << scanner` to right after `scanner.run!`

docs/configuration.md
```diff
@@ -46,6 +46,11 @@ project_name: "my-repo"
 # that might be used by the consumer of the Salus report.
 custom_info: "PR-123"
 
+# Numeric (Integer or Float)
+# Defines how many seconds Salus should run its scans before timing
+# out (defaults to indefinite amount of time if not specified)
+salus_timeout_s: 120
```
ghbren, line 52: My guess is that this `salus_timeout_s` will override `scanner_timeout_s`?
If so, could you clarify this in the doc?

Please test out using `scanner_timeout_s` and `salus_timeout_s` in the same salus.config to ensure there are no strange issues.
Please use `docker build ...` then `docker run ...` to test this, and test different combinations of `salus_timeout_s` and `scanner_timeout_s`, something like `salus_timeout_s` < `scanner_timeout_s` and  `salus_timeout_s` > `scanner_timeout_s`

ghbren:
Looks like we need to support another config option in this PR, something like `fail_on_salus_timeout`

If timed out:
- If `fail_on_salus_timeout=false`, then nothing needed
- If `fail_on_salus_timeout=true`, then set overall salus status to fail.  My guess is that you can add a variable like `salus_timedout` to `report.rb`.    Then in `report.rb` `def passed`, add something at the beginning, like  `return false if salus_timed_out`.

After you implement this, please test with `docker build...` and `docker run ...`

And please add doc for this.

PR #415, opened 2021-07-30
CycloneDX Validation

jeffrey778zhan:
**What Changed?** 
Added support to validate CycloneDX generated reports against json-schema.

**How was it tested?**
Unit tests

**Usage of CycloneDX Schema**
Usage of CycloneDX Json Schema approved


lib/cyclonedx/schema/bom-1.3.schema.json
```diff
@@ -0,0 +1,1054 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
```
jeffrey778zhan, line 2: The [schema](https://github.com/CycloneDX/specification/blob/master/schema/bom-1.3.schema.json) provided uses draft-07 compared to draft-04 which is the maximum version our `json-schema` gem supports. Potentially in another PR we'd like to support draft-04 and higher with this [gem](https://github.com/davishmcclurg/json_schemer).

lib/cyclonedx/report.rb
```diff
@@ -28,13 +33,22 @@ def to_cyclonedx
         metadata: {},
         components: []
       }
-      cyclonedx_report[:metadata] = converter.build_metadata
 
       # for each scanner report, run the appropriate converter
       @scan_reports.each do |scan_report|
-        cyclonedx_report[:components] << converter(scan_report[0])
+        cyclonedx_report[:components] += converter(scan_report[0])
```
joshua-ostrom, line 39: Did convertor go from returning an item to returning an array?

jeffrey778zhan: Right, converter returns an array of components and the final CycloneDX report expects a list of components rather than a list of lists so had to make this change.

lib/cyclonedx/base.rb
```diff
@@ -1,31 +1,21 @@
...
-          "purl": ""
-        }
-      }
-    end
-
     # Returns the 'components' object for a supported/unsupported scanner's report
     def build_components_object
-      components = []
       info = @scan_report.to_h.fetch(:info)
+      return [] unless info[:dependencies]
```
ghbren, line 15: If there are no dependencies, would `info[:dependencies]` be `nil` or something like `[]`?

lib/cyclonedx/report.rb
```diff
@@ -9,13 +11,16 @@
...
+    class CycloneDXInvalidFormatError < StandardError; end
+
     def initialize(scan_reports, config = {})
       @scan_reports = scan_reports
       @config = config
     end
 
     CYCLONEDX_SPEC_VERSION = "1.3".freeze
-    CYCLONEDX_VERSION = "1".freeze
+    CYCLONEDX_VERSION = 1
```
ghbren, line 23: Why was the `freeze` taken out?

jeffrey778zhan: Rubocop said it was redundant `Style/RedundantFreeze: Do not freeze immutable objects, as freezing them has no effect.`

lib/cyclonedx/base.rb
jeffrey778zhan: Yep `info[:dependencies]` becomes nil which causes errors if the code execution proceeds to ` info[:dependencies].each` leading to an exception

ghbren (approved):

PR #417, opened 2021-08-05
Salus verb header support in config yaml

jeffrey778zhan:
**What's changed?**
- Added Salus support for CycloneDX-json format as well as new headers/environment_headers and verbs (put, post only for now, we can add as we need). Changed Docs to reflect that. 
- Refactored logic in `report.rb` specific to creating an HTTP request i.e the header and faraday to another class `ReportRequest`
- Added in cyclonedx_options support in config yaml file. More details in docs. 

**How has it been tested?**
Unit tests and tested locally with simple server that echoes back requests.


lib/salus/report.rb
```diff
@@ -205,11 +213,15 @@ def publish_report(directive)
                       when 'yaml' then to_yaml
                       when 'sarif' then to_sarif(directive['sarif_options'] || {})
                       when 'sarif_diff' then to_sarif_diff
+                      when 'cyclonedx-json' then to_cyclonedx(directive['cyclonedx_options'] || {})
                       else
                         raise ExportReportError, "unknown report format #{directive['format']}"
                       end
       if Salus::Config::REMOTE_URI_SCHEME_REGEX.match?(URI(uri).scheme)
-        send_report(uri, report_body(directive), directive['format'])
+        send_report(directive,
```
jeffrey778zhan, line 221: Could simply pass in directive and read `directive['format']` from there as well.

lib/salus/report.rb
```diff
@@ -334,7 +373,12 @@ def report_body(config)
         body = report_body_hash(config, JSON.parse(to_sarif(config['sarif_options'] || {})))
       end
       body = report_body_hash(config, JSON.parse(to_sarif_diff)) if config['format'] == 'sarif_diff'
-      return JSON.pretty_generate(body) if %w[json sarif sarif_diff].include?(config['format'])
+      if config['format'] == 'cyclonedx-json'
+        body = report_body_hash(config, JSON.parse(to_cyclonedx(config['cyclonedx_options'] || {})))
```
jeffrey778zhan, line 377: Want to add in config['cyclonedx_options'] support in a future PR for CycloneDX Spec versions.

lib/salus/report.rb
```diff
@@ -189,7 +190,14 @@ def get_sarif_results(sarif_data)
     end
 
     def to_cyclonedx(config = {})
-      Cyclonedx::Report.new(@scan_reports, config).to_cyclonedx
+      cyclonedx_bom = Cyclonedx::Report.new(@scan_reports, config).to_cyclonedx
+      cyclonedx_report = {
+        autoCreate: true,
+        projectName: "#{@builds['org']}/#{@builds['project']}",
```
jeffrey778zhan, line 196: Capture projects at `{org/project}` level

ghbren, line 196: `@builds` will only have value value if `builds` is defined in the salus config.  
If `builds` is not set, the code above will set `projectName` to `/`.

Perhaps you should just set projectName to `@project_name`.

docs/configuration.md
```diff
@@ -54,13 +54,20 @@ custom_info: "PR-123"
...
 # 
 # Each report hash can add post paramaters using the `post` key , 
 # - Salus reports can be sent as a report paramater by specifying the paramater name in `salus_report_param_name`
 # - additional post paramaters can be specified through the `additional_params` field
 #
+# Each report hash can also specify what http verb should be used (currently support `put` and `post` key),
+# and headers to set. 
+# - If not specified http verb defaults to `post`
+# - you can pass in `headers` or `headers_env_var` with the corresponding list of name and value pairs.
+# - Values specified in `headers_env_var` will be treated as an environment variable `ENV[val]` while values
```
joshua-ostrom, line 68: Salus has a pattern in place for ENV vars:
https://github.com/coinbase/salus/blob/8de754473f8efbc19574e04152a921c8a7644497/lib/salus/config.rb#L66
https://github.com/coinbase/salus/blob/8de754473f8efbc19574e04152a921c8a7644497/lib/salus/config.rb#L171-L187
https://github.com/coinbase/salus/blob/61505620672bdabb486e8454b9dd3867365a50da/spec/fixtures/config/envar_config.yaml#L2-L10

You should be able to leverage that here

docs/configuration.md
```diff
@@ -79,6 +86,13 @@ reports:
       additional_params:
         repo: 'Random Repo'
         user: 'John Doe' 
+  - uri: https://salus-config.internal2.net/salus-report
+    format: cyclonedx-json
+    put:
+    headers:
+      Age: '12'
+    headers_env_var:
```
joshua-ostrom, line 94: See note above this can go away

lib/salus/report.rb
joshua-ostrom: I might suggest you make this a configuration option.  That would allow the user to specify any value or combination of fields.  

For example:
```
  - uri: https://salus-config.internal2.net/salus-report
    format: cyclonedx-json
    put:
     cyclone_project_name:  "{{SALUS_BUILD_ORG}}/{{SALUS_BUILD_PROJECT}}"
      headers:
        Age: '12'
```

joshua-ostrom: Oh I see cyclonedx_options below... perfect place for this field IMHO

lib/salus/report.rb
```diff
@@ -301,22 +313,49 @@ def write_report_to_file(report_file_path, report_string)
...
       unless response.success?
         raise ExportReportError,
-              "POST of Salus report to #{remote_uri} had response status #{response.status}."
+              "Salus report to #{remote_uri} had response status #{response.status}."
       end
     end
 
     def report_body_hash(config, data)
-      return data unless config&.key?('post')
+      return data unless config&.key?('post') && !config['post'].nil?
```
joshua-ostrom, line 358: config['key'].present?

Salus has activesupport in the gemfile so you can take advantage of some of that syntactical sugar

lib/salus/report.rb
```diff
@@ -301,22 +313,49 @@ def write_report_to_file(report_file_path, report_string)
...
       end
 
+      header_hash
+    end
+
+    def send_report(config, remote_uri, data, format)
+      conn = Faraday.new(
+        url: remote_uri,
+        headers: report_headers_h(config['headers'] || {},
+                                  config['headers_env_var'] || {},
```
joshua-ostrom, line 337: this can go away

lib/salus/report.rb
```diff
@@ -301,22 +313,49 @@ def write_report_to_file(report_file_path, report_string)
             "Cannot write file #{report_file_path} - #{e.class}: #{e.message}"
     end
 
-    def send_report(remote_uri, data, format)
-      response = Faraday.post do |req|
-        req.url remote_uri
-        req.headers['Content-Type'] = CONTENT_TYPE_FOR_FORMAT[format]
-        req.headers['X-Scanner'] = x_scanner_type(format)
-        req.body = data
+    def report_headers_h(headers, headers_env_var, format)
```
joshua-ostrom, line 316: Might be a win to move this to a dedicated class.  report.rb is growing rather large and ripe for some refactoring.

lib/salus/report.rb
```diff
@@ -301,22 +313,49 @@ def write_report_to_file(report_file_path, report_string)
...
+      header_hash = {}
+      header_hash['Content-Type'] = CONTENT_TYPE_FOR_FORMAT[format]
+      header_hash['X-Scanner'] = x_scanner_type(format)
+
+      # Iterate through specified headers and assign values
+      headers.each do |field, value|
```
joshua-ostrom, line 322: header_hash.merge!(headers)

spec/lib/salus/processor_spec.rb
```diff
@@ -215,5 +227,64 @@ def remove_key(json_string, key = 'running_time')
...
+          remote_uri1,
+          headers:
+            {
+              'Content-Type' => 'application/json',
+              'X-API-Key' => '',
+              'repo' => 'Random Repo'
+            },
+          times: 1
+        ) do |req|
+          expect(req.body).to match_cyclonedx_report_json(expected_report)
```
joshua-ostrom, line 285: Are there specific fields in the bom that require you to drop the bom from the cyclonedx match check here?

lib/salus/report.rb
jeffrey778zhan: By builds being defined in the salus config do you mean that the salus config contains the build param or if the {{SALUS_BUILD_ORG}} and {{SALUS_BUILD_PROJECT}} are set? I think in our case since we will be changing the "default" Salus scheduled scan yaml we'll be able to guarantee that builds params is set. My worry with using @project_name is that it specifies the project down to the commit level which will create a lot of outdated entries leading to false alarms for old code. Maybe we can use project name org and project are empty.

spec/lib/salus/processor_spec.rb
jeffrey778zhan: Yep, the UUID field causes the base64 encoding to be different. I may be able to decode, remove that field and encode again 🤔

lib/salus/report.rb
jeffrey778zhan: Refactored logic in report.rb specific to creating an HTTP request i.e the header and faraday to another class ReportRequest

lib/salus/report.rb
jeffrey778zhan: @ghbren @joshua-ostrom Added in support for `cyclonedx_project_name` option in `cyclonedx_options`. If empty I default to empty string. Brianna mentioned that we could also default to `project_name` config var but users may want the flexibility to differentiate `project_name` from cyclonedx and have the cyclone project name as empty so left that out for now.

lib/salus/report_request.rb
```diff
@@ -0,0 +1,55 @@
...
+
+        response = if config&.key?('put')
+                     conn.put do |req|
+                       req.body = data
+                     end
+                   else
+                     conn.post do |req|
+                       req.body = data
+                     end
+                   end
```
joshua-ostrom, line 31: you might be able to slim this down to:

```
verb = config&.key?('put') ? :put : :post
conn.send(verb) { |req| req.body = data }```

File: lib/salus/report_request.rb [hunk 12]
```diff
@@ -0,0 +1,55 @@
...
+
+      def report_headers_h(headers, format)
+        header_hash = {}
+        header_hash['Content-Type'] = CONTENT_TYPE_FOR_FORMAT[format]
+        header_hash['X-Scanner'] = x_scanner_type(format)
+        header_hash.merge!(headers)
+      end
+
+      def x_scanner_type(format)
+        if format == 'sarif_diff'
```
joshua-ostrom, line 47: minor nit... would prefer consts:
```
return SCANNER_TYPE_SARIF_DIFF if format == FORMAT_SARIF_DIFF
SCANNER_TYPE_SALUS
```

lib/salus/report_request.rb
```diff
@@ -0,0 +1,55 @@
+require 'faraday'
+
+module Salus
```
joshua-ostrom, line 3: Nice job refactoring this out 👍

joshua-ostrom (approved):

ghbren (approved):

PR #419, opened 2021-08-09
Add Support for Rust CycloneDx

saron-deb:
Adding support for Rust for CycloneDx implementation


spec/lib/cyclonedx/report_rust_crates_cyclonedx_spec.rb
```diff
@@ -0,0 +1,213 @@
...
+        mock_lock = File.join(repo.path_to_repo, 'Cargo.lock')
+        FileUtils.cp existing_lock, mock_lock
+      end
+
+      scanner.run
+
+      rust_cyclonedx = Cyclonedx::ReportRustCrates.new(scanner.report)
+      expected = [
+        {
+          "bom-ref": "pkg:cargo_lock/autocfg@0.1.7",
```
saron-deb, line 89: `"pkg:cargo_lock/autocfg@0.1.7"`

Unsure if "cargo_lock" is the desired type here? Simple to change in report_rust_crates.rb if necessary.

ghbren: Is `cargo_lock` the only `type` for Rust?

lib/cyclonedx/report_rust_crates_cyclonedx.rb
```diff
@@ -3,5 +3,14 @@ class ReportRustCrates < Base
...
       super(scan_report)
     end
+
+    def package_url(dependency)
+      "pkg:#{dependency[:type]}/#{dependency[:name]}@#{version_string(dependency)}"
+    end
+
+    # Return version string to be used in purl or component
+    def version_string(dependency)
+      (dependency[:version_tag]).to_s
```
jeffrey778zhan, line 13: Will the version tag always be `@x.x.x` ?

lib/cyclonedx/base.rb
```diff
@@ -33,7 +33,8 @@ def parse_dependency(dependency)
         "properties": [
           {
             "key": "source",
-            "value": dependency[:source]
+            # Varies between these two values by scanner
+            "value": dependency[:source] || dependency[:reference]
```
joshua-ostrom, line 37: Aside: We probably should consider standardizing the existing dependency reporters.  This would allow you to only look for source here.  Not something we need to hit in this PR - just making a note.

lib/cyclonedx/report_rust_crates_cyclonedx.rb
joshua-ostrom: minor nit you don't need the () here

saron-deb: I think so it seems like rust follows semantic versioning 

 http://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/cargo/reference/manifest.html#:~:text=Use%20version%20numbers%20with%20three%20numeric%20parts%20such%20as%201.0.0%20rather%20than%201.0.

spec/lib/cyclonedx/report_rust_crates_cyclonedx_spec.rb
```diff
@@ -0,0 +1,213 @@
...
+        ]
+      )
+    end
+
+    it 'should calculate and report all the dependencies in the manifest if Cargo.lock is absent' do
+      repo = Salus::Repo.new('spec/fixtures/report_rust_crates/manifest_only')
+      scanner = Salus::Scanners::ReportRustCrates.new(repository: repo, config: {})
+
+      # We will stub the lock file generate to keep our specs from needing to
+      # hit the internet and pulldown the dependency graph
```
joshua-ostrom, line 77: Thanks for the comment 💯 Maybe add to it that cargo tree will create a lock file if not already present for additional clarity

spec/lib/cyclonedx/report_rust_crates_cyclonedx_spec.rb
saron-deb: In report_rust_crates.rb, it seems we're currently extrapolating all dependency info from cargo.lock if it's available or generating it if it's not. 

The type field is hard coded in report_rust_crates.rb to `cargo_lock` similar to other scanners, I've reached out to Maddie about this in general to see what implications it could have w/Dep track

saron-deb: Maddie got back to me and let me know that the type in the purl field is sent by Dep track to other programs to get more data. 

I'll change this to match list of defined types @ https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst

and do likewise for other scanners 👍

joshua-ostrom (dismissed):

joshua-ostrom (approved):

PR #420, opened 2021-08-09
Adding support for Python CycloneDx

saron-deb:
Adding support for Python in CycloneDx implementation

### DIFF
--- a/lib/cyclonedx/base.rb
+++ b/lib/cyclonedx/base.rb
@@ -33,7 +33,7 @@ module Cyclonedx
         "properties": [
           {
             "key": "source",
-            "value": dependency[:source] || "" # In case unavailable for scanner
+            "value": dependency[:source].to_s
           },
           {
             "key": "dependency_file",
--- a/lib/cyclonedx/base.rb
+++ b/lib/cyclonedx/base.rb
@@ -33,7 +33,7 @@ module Cyclonedx
         "properties": [
           {
             "key": "source",
-            "value": dependency[:source]
+            "value": dependency[:source] || "" # In case unavailable for scanner
           },
           {
             "key": "dependency_file",
--- a/lib/cyclonedx/report_python_modules_cyclonedx.rb
+++ b/lib/cyclonedx/report_python_modules_cyclonedx.rb
@@ -3,5 +3,19 @@ module Cyclonedx
     def initialize(scan_report)
       super(scan_report)
     end
+
+    def package_url(dependency)
+      "pkg:#{dependency[:type]}/#{dependency[:name]}#{version_string(dependency, true)}"
+    end
+
+    # Return version string to be used in purl or component
+    def version_string(dependency, is_purl_version = false)
+      # If the dependency is specified in requirements.txt and an absolute version is needed for
+      # the purl return empty
+      return "" if dependency[:dependency_file] == 'requirements.txt' && is_purl_version
+
+      prefix = is_purl_version ? "@" : ""
+      "#{prefix}#{dependency[:version]}"
+    end
   end
 end
--- a/lib/cyclonedx/report_python_modules_cyclonedx.rb
+++ b/lib/cyclonedx/report_python_modules_cyclonedx.rb
@@ -10,11 +10,8 @@ module Cyclonedx
 
     # Return version string to be used in purl or component
     def version_string(dependency, is_purl_version = false)
-      # Duplicate for the second time this is called in base.rb for the purl field
-      # Prevents any issues with the initial gsub for the first call
-      dummy_version = dependency[:version].dup
       # Check if dependency is pinned
-      is_pinned = !!dummy_version.match(/^==[0-9]/)
+      is_pinned = !!dependency[:version].match(/^==[0-9]/)
 
       # If unpinned dependency is specified in requirements.txt
       # and version is needed for purl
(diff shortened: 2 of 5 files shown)


spec/lib/cyclonedx/report_python_modules_cyclonedx_spec.rb
```diff
@@ -0,0 +1,255 @@
...
+
+      python_cyclonedx = Cyclonedx::ReportPythonModules.new(scanner.report)
+      expected = [
+        {
+          "bom-ref": "pkg:python_requirement/requests",
+          "type": "library",
+          "group": "",
+          "name": "requests",
+          "version": ">=2.5",
+          "purl": "pkg:python_requirement/requests",
```
saron-deb, line 19: `"purl": "pkg:python_requirement/requests"`

Unsure if "python_requirement" is the desired type here? Simple to change in report_python_modules.rb if necessary.

ghbren: Is `python_requirement` the only possible type?

spec/lib/cyclonedx/report_python_modules_cyclonedx_spec.rb
```diff
@@ -0,0 +1,255 @@
...
+      scanner.run
+
+      python_cyclonedx = Cyclonedx::ReportPythonModules.new(scanner.report)
+      expected = [
+        {
+          "bom-ref": "pkg:python_requirement/amqp",
+          "type": "library",
+          "group": "",
+          "name": "amqp",
+          "version": "==2.2.2",
```
ghbren, line 83: Maybe remove `==` so that we have this?
```json
"version": "2.2.2",
```
What do you think, Josh?

lib/cyclonedx/report_python_modules_cyclonedx.rb
```diff
@@ -3,5 +3,19 @@ class ReportPythonModules < Base
...
+
+    def package_url(dependency)
+      "pkg:#{dependency[:type]}/#{dependency[:name]}#{version_string(dependency, true)}"
+    end
+
+    # Return version string to be used in purl or component
+    def version_string(dependency, is_purl_version = false)
+      # If the dependency is specified in requirements.txt and an absolute version is needed for
+      # the purl return empty
+      return "" if dependency[:dependency_file] == 'requirements.txt' && is_purl_version
```
jeffrey778zhan, line 15: Seems like python doesn't use a lock file but can specify exact versions as show https://iscompatible.readthedocs.io/en/latest/ruby%20encode through here through the `==`. We could aim to include the one's with pinned or semi pinned versions in the purl

spec/lib/cyclonedx/report_python_modules_cyclonedx_spec.rb
saron-deb: I think the requirements.txt is the only file we're gathering python dependencies from, but that type is currently hardcoded in report_python_modules.rb (https://github.com/coinbase/salus/blob/03279595de981a0aaad278f90f51d797e728e348/lib/salus/scanners/report_python_modules.rb#L20)

Maybe something else would be more suitable than `python_requirement`?

spec/lib/cyclonedx/report_python_modules_cyclonedx_spec.rb
saron-deb: Yep I can remove `==` for pinned requirements. It's just for unpinned/semi-pinned requirements, if something like 

`>=3.2.1` 

is given for example, getting rid of `>=` might cause confusion as we wouldn't be able to tell that any newer version (3.2.1+) is also specified?

lib/cyclonedx/report_python_modules_cyclonedx.rb
jeffrey778zhan: Yep!

lib/cyclonedx/base.rb
```diff
@@ -33,7 +33,7 @@ def parse_dependency(dependency)
         "properties": [
           {
             "key": "source",
-            "value": dependency[:source]
+            "value": dependency[:source] || "" # In case unavailable for scanner
```
joshua-ostrom, line 36: dependency[:source].to_s should do the same thing

lib/cyclonedx/report_python_modules_cyclonedx.rb
```diff
@@ -3,5 +3,34 @@ class ReportPythonModules < Base
...
+
+    def package_url(dependency)
+      "pkg:#{dependency[:type]}/#{dependency[:name]}#{version_string(dependency, true)}"
+    end
+
+    # Return version string to be used in purl or component
+    def version_string(dependency, is_purl_version = false)
+      # Duplicate for the second time this is called in base.rb for the purl field
+      # Prevents any issues with the initial gsub for the first call
+      dummy_version = dependency[:version].dup
```
joshua-ostrom, line 15: can you just use gsub in place of gsub! ?

lib/cyclonedx/report_python_modules_cyclonedx.rb
```diff
@@ -3,5 +3,34 @@ class ReportPythonModules < Base
...
+
+      # If unpinned dependency is specified in requirements.txt
+      # and version is needed for purl
+      if dependency[:dependency_file] == 'requirements.txt' && is_purl_version && !is_pinned
+        return ""
+      end
+
+      prefix = is_purl_version ? "@" : ""
+
+      # Will only gsub if absolute/pinned dependency version
```
joshua-ostrom, line 27:
```
return "#{prefix}#{dependency[:version]}" unless is_pinned

"#{prefix}#{dummy_version.gsub(/^==/, '')}"
```

joshua-ostrom (approved):

PR #421, opened 2021-08-16
CycloneDX Node Modules Support With Purl Percentage Encoding

jeffrey778zhan:
**What changed?**
- Added in percentage encoding for purl following the [spec](https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst#how-to-build-purl-string-from-its-components). Currently support type, namespace and version (not including qualifiers and subpaths). You can test if a purl can be recognized by using the search option for this vulnerability scanner https://ossindex.sonatype.org/
- Added in logic to parse node module dependencies and populate cyclonedx reports

**How was it tested?**
- Specs for purl encoding
- Testing valid purl's through https://ossindex.sonatype.org/
- Specs for node modules package.json, package-json.lock and yarn.lock as sources

### DIFF
--- a/lib/cyclonedx/package_url.rb
+++ b/lib/cyclonedx/package_url.rb
@@ -1,0 +1,30 @@
+module Cyclonedx
+  class PackageUrl
+    def initialize(type:, namespace:, version:)
+      @type = type
+      @namespace = namespace
+      @version = version
+    end
+
+    # TODO: Add support for qualifiers and subpaths
+    # https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst#how-to-build-purl-string-from-its-components
+    def to_string
+      # Start with type and a colon
+      purl = "pkg:#{@type}/"
+
+      # Add the required namespace
+      if @namespace.present?
+        # Remove leading and trailing /
+        namespace = @namespace.delete_prefix("/").delete_suffix("/")
+
+        ns = namespace.split("/").map { |s| CGI.escape(s) }
+        purl += ns.join("/")
+      end
+
+      # If a version is provided, add it after the @ symbol
+      purl += "@#{CGI.escape(@version)}" if @version.present?
+
+      purl
+    end
+  end
+end
--- a/lib/cyclonedx/report_node_modules_cyclonedx.rb
+++ b/lib/cyclonedx/report_node_modules_cyclonedx.rb
@@ -1,18 +1,11 @@
 module Cyclonedx
   class ReportNodeModules < Base
     TYPE = "npm".freeze
+    UNLOCKED_DEPENDENCY_FILE = "package.json".freeze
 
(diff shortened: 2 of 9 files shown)


lib/cyclonedx/report_node_modules_cyclonedx.rb
```diff
@@ -1,7 +1,18 @@
 module Cyclonedx
   class ReportNodeModules < Base
+    TYPE = "npm".freeze
```
jeffrey778zhan, line 3: @saron-deb If you feel that Python, Rust, and Go can also fit this style it could be a win to cut down some code.

saron-deb: Yep they definitely can, will make that change (in a separate PR?)

jeffrey778zhan: Yep separate PR once this one is merged 👍

lib/salus/scanners/report_node_modules.rb
```diff
@@ -136,12 +136,10 @@ def record_dependencies_from_package_json(record_modules: true)
         # Resolve the source if possible from .npmrc
         source = if name.start_with?('@')
                    sections = name.split(%r{\/})
-                   name = sections[1]
```
jeffrey778zhan, line 139: Previously we would be removing the prefix of the name in how we were parsing names that started with @. E.g `@nerv-hq/control-system` would become `control-system`. If there wasn't any reasoning for the previous logic the additional information in the name could be useful.

lib/cyclonedx/report_ruby_gems_cyclonedx.rb
```diff
@@ -1,21 +1,18 @@
...
     def initialize(scan_report)
       super(scan_report)
     end
 
-    def package_url(dependency)
-      "pkg:#{dependency[:type]}/#{dependency[:name]}#{version_string(dependency, true)}"
-    end
-
     # Return version string to be used in purl or component
     def version_string(dependency, is_purl_version = false)
```
joshua-ostrom, line 10: This looks like the same logic as the def version_string above... you may wanna DRY this up

lib/cyclonedx/package_url.rb
```diff
@@ -0,0 +1,30 @@
+module Cyclonedx
```
joshua-ostrom, line 1: Nice decomposition here

lib/cyclonedx/report_ruby_gems_cyclonedx.rb
jeffrey778zhan: Agreed, made a change to refactor the `version_string` method to the super class

joshua-ostrom (dismissed):

joshua-ostrom (approved):

PR #425, opened 2021-08-18
Add a SARIF Adapter for RepoNotEmpty

cbged2:


spec/lib/sarif/repo_not_empty_sarif_spec.rb
```diff
@@ -0,0 +1,58 @@
+require_relative '../../spec_helper'
+
+describe Sarif::RepoNotEmptySarif do
+  describe '#build_runs_object' do
+    let(:scanner) { Salus::Scanners::RepoNotEmpty.new(repository: repo, config: {}) }
+    before { scanner.run }
+
+    context 'vulnerabilites found in project' do
```
joshua-ostrom, line 8: Can you make a context for a non-empty project too - that will allow us to test the common case

cbged2:
@joshua-ostrom Please see #428. I've pushed updated changes from that account and will be using that from now on.

PR #428, opened 2021-08-20
Add a SARIF Adapter for RepoNotEmpty

geraldnash-cb:
Opening this in favor of #425, as this account is part of the Coinbase org.


geraldnash-cb:
@joshua-ostrom

joshua-ostrom (dismissed):

PR #430, opened 2021-08-23
CycloneDx Golang Support

saron-deb:
Adding support for Golang in CycloneDx implementation.

### DIFF
--- a/lib/cyclonedx/report_go_dep_cyclonedx.rb
+++ b/lib/cyclonedx/report_go_dep_cyclonedx.rb
@@ -1,7 +1,21 @@
 module Cyclonedx
   class ReportGoDep < Base
+    TYPE = "golang".freeze
+    UNLOCKED_DEPENDENCY_FILE = "go.sum".freeze
     def initialize(scan_report)
       super(scan_report)
+    end
+
+    # Return version string to be used in purl or component
+    def version_string(dependency, is_purl_version = false)
+      # If the dependency is specified in a unlocked dependency file and an absolute version
+      # is needed for the purl return empty
+      if (dependency[:dependency_file] == self.class::UNLOCKED_DEPENDENCY_FILE ||
+          dependency[:dependency_file] == "Gopkg.lock") && is_purl_version
+        return ""
+      end
+
+      dependency[:version]
     end
   end
 end
--- a/lib/cyclonedx/report_go_dep_cyclonedx.rb
+++ b/lib/cyclonedx/report_go_dep_cyclonedx.rb
@@ -1,21 +1,9 @@
 module Cyclonedx
   class ReportGoDep < Base
     TYPE = "golang".freeze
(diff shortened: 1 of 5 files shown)


spec/lib/cyclonedx/report_go_dep_cyclonedx_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+
+      golang_cyclonedx = Cyclonedx::ReportGoDep.new(scanner.report)
+      expect(golang_cyclonedx.build_components_object).to match_array(
+        [
+          {
+            "bom-ref": "pkg:golang/github.com/PagerDuty/go-pagerduty",
+            "type": "library",
+            "group": "",
+            "name": "github.com/PagerDuty/go-pagerduty",
+            "version": nil,
```
jeffrey778zhan, line 85: I think setting nil may result in an invalid cyclonedx report, can you add in a test to `cyclonedx_report_spec` similar to https://github.com/coinbase/salus/blob/master/spec/lib/cyclonedx/cyclonedx_report_spec.rb#L13 to test if this will raise an error.

lib/cyclonedx/report_go_dep_cyclonedx.rb
```diff
@@ -1,7 +1,21 @@
 module Cyclonedx
   class ReportGoDep < Base
+    TYPE = "golang".freeze
+    UNLOCKED_DEPENDENCY_FILE = "go.sum".freeze
```
jeffrey778zhan, line 4: I think `go.mod` should be set here since go.sum contains the locked versions of the dependency? Otherwise the version of dependencies won't be included in the purl

spec/lib/cyclonedx/report_go_dep_cyclonedx_spec.rb
saron-deb: will do!

lib/salus/scanners/report_go_dep.rb
```diff
@@ -36,7 +36,7 @@ def record_dep_from_go_sum
             {
               "namespace" => (matches[:namespace]).to_s,
               "name" => (matches[:name]).to_s,
-              "version" => (matches[:version]).to_s.gsub(%r{/go.mod}, '').strip,
+              "version" => (matches[:version]).to_s.gsub(/^v/, '').gsub(%r{/go.mod}, '').strip,
```
joshua-ostrom, line 39: I assume dropping v from the start of the versions won't impact anything with dependency track?

jeffrey778zhan: Would be nice to double check with the search option here https://ossindex.sonatype.org/

saron-deb: Will revert just to be completely safe 👍

joshua-ostrom (approved):

PR #404, opened 2021-07-22
Update Semgrep to 0.59.0, disable version check

zbuc:
This updates Semgrep to 0.59.0 which has some important grammar fixes to various languages. This PR also disables the semgrep version check, as the version check output goes to STDERR for some semgrep invocations (specifically I noticed it when putting semgrep rules in an external file), and can cause issues in CI pipelines that look at STDERR.

PR #431, opened 2021-08-23
Update semgrep to 0.62.0

zbuc:
Recreating https://github.com/coinbase/salus/pull/404 because CI didn't run correctly.

This updates Semgrep to 0.62.0 which has some important grammar fixes to various languages. This PR also disables the semgrep version check, as the version check output goes to STDERR for some semgrep invocations (specifically I noticed it when putting semgrep rules in an external file), and can cause issues in CI pipelines that look at STDERR.


ghbren:
0.62.0 added to salus in another PR

PR #433, opened 2021-08-24
Let Salus Report SARIF Formats to Remote Destinations

geraldnash-cb:
Right now Salus doesn't define its `X-Scanner` header to support normal SARIF formats. In order for entities that receive reports over HTTP to know when they're receiving SARIF reports, Salus should indicate when it's using this format.


geraldnash-cb:
@ghbren

ghbren (approved):

PR #437, opened 2021-08-31
Add dependency_of to SARIF result and update the schema

hesh64:


lib/sarif/schema/sarif-schema.json
```diff
@@ -1280,6 +1280,10 @@
           "description": "A plain text message string.",
           "type": "string"
         },
+        "dependency_of": {
+          "description": "The name of the package that depends on this package",
+          "type": "string"
```
hesh64, line 1285: we can skip the schema change by appending dep_of to the message text

PR #441, opened 2021-09-06
Fix skips config name for Bandit

hstocks:
The [Salus documentation](https://github.com/coinbase/salus/blob/master/docs/scanners/bandit.md) for the Bandit configuration specifies that rules to skip should be listed in YAML with `skips`,  which matches [Bandit's documentation](https://bandit.readthedocs.io/en/latest/config.html), but the code actually uses `skip`.  This commit fixes that.


ghbren:
Is this PR ready for review?  If so, could you fix the failed circle ci tests?

PR #442, opened 2021-09-06
Add more Python project detections

hstocks:
This adds `setup.py` (for packages), `environment.yml` and `environment.yaml` (for Conda) as indicators that Bandit should run.


ghbren:
@hstocks , could you fix the rubocop failures?  Thanks

PR #443, opened 2021-09-15
improve brakeman sarif and fix fingerprint overwrite

ghbren:
Updated brakeman sarif adapter 
* Removed `fingerprint` from `messageStrings`.  Note multiple vulnerabilities could have the same code but different fingerprints.  But `messageStrings` was added to `rules`, where we could only keep one fingerprint for all vulnerabilities that have the same code.
* Removed `confidence` from `messageStrings`, because there could be multiple confidences for vulnerabilities that have the same code.
  - We should look into this issue for other adapters as future work.
* Added a `properties` field that gets added to `results`, with
  - `fingerprint`
  - `confidence`
  - `render_path`
  - `user_input`
  - `location_type`
  - `location_class`
  - `location_method`


joshua-ostrom (approved):

joshua-ostrom (approved):

geraldnash-cb (approved):

spec/lib/sarif/brakeman_sarif_spec.rb
```diff
@@ -19,17 +19,23 @@
...
           start_column: 1,
           help_url: "https://brakemanscanner.org/docs/warning_types/dangerous_eval/",
           code: "eval(params[:evil])",
-          uri: "app/controllers/static_controller_controller.rb"
+          uri: "app/controllers/static_controller_controller.rb",
+          properties: {
+            fingerprint: "b16e1cd0d952433f80b0403b6a74aab0e98792ea015cc1b1fa5c003cbe7d56eb",
+            confidence: "High",
+            render_path: "",
+            user_input: "params[:evil]",
```
hesh64, line 34: Evil is lurking

hesh64 (approved):

PR #444, opened 2021-09-16
upgrade salus container to use go 1.16.8

CoinsNaulty:
GoSec release notes indicate that it works with 1.16.x golang versions.

This commit updates the golang version to `1.16.8`.
This should reduce errors when scanning golang source code using `go
1.16`

Example Error without this change:
```
"Golang errors": {
      "api//home/repo/api/api.go": [
        {
          "line": 78,
          "column": 17,
          "error": "ReadAll not declared by package io"
        }
```


CoinsNaulty:
relates to: https://github.com/coinbase/salus/pull/365

CoinsNaulty:
Implemented an alternative solution here: https://github.com/bisontrails/salus/pull/1/commits/40ecc90e8060b49450ec202abb1e188a3a5270f7
This gives us more flexibility when running salus on various projects that have different golang version requirements.

PR #449, opened 2021-09-20
SECT 912 Normalize bandit exceptions

joshua-ostrom:
Update bandit exceptions to use the new normalized exception format.


ghbren:
LGTM!

PR #450, opened 2021-09-20
Sect 912 Timebox Brakeman

joshua-ostrom:
Normalize the exception form at for brakeman and support timeboxing.


ghbren:
circle tests failed

lib/salus/scanners/brakeman.rb
```diff
@@ -50,11 +66,65 @@ def self.supported_languages
       ['ruby']
     end
 
+    def run_with_exceptions_applied
+      # create a temporary file combining ignore file entries with any user supplied
+      # entires if exceptions hash is being used
+
+      Tempfile.create('salus') do |f|
```
ghbren, line 73: Did you test this with `docker run ...`?
I could be wrong, but it seems like docker is read only by default, and we can't write things unless some flag is set.  If that's the case, maybe we'll want to include this in the doc or something?

joshua-ostrom: Good call out, included spec coverage and error reporting

ghbren (approved):

PR #453, opened 2021-09-20
Sect 912 Normalize GoSec Exceptions

joshua-ostrom:
Update gosec exception config to support the normalized exception format.  This will enable timeboxing as well.

### DIFF
--- a/lib/salus/scanners/gosec.rb
+++ b/lib/salus/scanners/gosec.rb
@@ -190,8 +190,13 @@ module Salus::Scanners
           # exlude the folders from scan
           # can be files or directories
           'exclude-dir': :file
-        }
+        },
+        config_overrides: { 'exclude' => excluded_ids }
       )
+    end
+
+    def excluded_ids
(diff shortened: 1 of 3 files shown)


ghbren:
LGTM!

lib/salus/scanners/gosec.rb
```diff
@@ -190,10 +190,15 @@ def config_options
...
           # can be files or directories
           'exclude-dir': :file
-        }
+        },
+        config_overrides: { 'exclude' => excluded_ids }
       )
     end
 
+    def excluded_ids
+      (fetch_exception_ids + @config.fetch('exclude', [])).uniq
```
ghbren, line 199: is there a reason to also `sort`?

ghbren (approved):

lib/salus/scanners/gosec.rb
joshua-ostrom: i don't think so we just pass the exceptions to gosec for exclusion

PR #465, opened 2021-10-25
Support for --sarif_diff_full

ghbren:
Support for `--sarif_diff_full`. The command below filters out vulnerabilities in `sarif_old.json` from `sarif_new.json`.

```docker run --rm -t -v $(pwd):/home/repo salus-local --sarif_diff_full sarif_new.json sarif_old.json```

Result can be written with `salus.yaml`
```yaml
reports:
  - uri: file://diff_sarif.json
    format: sarif_diff_full
```

Tested with
* new unit tests
* `docker run ...` with test files


lib/salus.rb
```diff
@@ -55,6 +58,19 @@ def scan(
                                        ignore_config_id: ignore_config_id,
                                        cli_scanners_to_run: only)
 
+      if sarif_diff_full != ""
```
joshua-ostrom, line 61: Minor nit... can we break out this logic to a method to keep the current scan method from growing too much.

`process_sarif_full_diff(sarif_diff_full) unless sarif_diff_full.empty?`

      
```

File: lib/salus.rb [hunk 2]
```diff
@@ -55,6 +58,19 @@ def scan(
...
+        begin
+          processor.create_full_sarif_diff(sarif_diff_full)
+        rescue StandardError => e
+          puts "Failed to get sarif diff #{e.inspect}"
+          system_exit(EXIT_FAILURE)
+        end
+
+        processor.report.report_uris.select! { |u| u['format'] == FULL_SARIF_DIFF_FORMAT }
```
joshua-ostrom, line 69: Do we need a new format?  Might be nice to just reuse "sarif" as the diff is producing valid sarif.  We can probably retire the old sarif diff format at some point.  Thoughts?

lib/salus/processor.rb
```diff
@@ -113,6 +113,31 @@ def scan_project
       end
     end
 
+    def create_full_sarif_diff(sarif_diff_full)
+      sarif_file_new, sarif_file_old = sarif_diff_full.split(" ")
```
joshua-ostrom, line 117: I'm okay with this... but curious does thor support passing them without manually wrapping in quotes?

spec/lib/salus_spec.rb
```diff
@@ -117,6 +117,34 @@
...
+          Salus.scan(quiet: true, repo_path: '.', sarif_diff_full: args)
+          diff_file = 'diff_1_2.json'
+          expect(File).to exist(diff_file)
+          diff_sarif = JSON.parse(File.read(diff_file))
+          expected_sarif = JSON.parse(File.read('diff_1_2.json'))
+          expect(expected_sarif).to eq(diff_sarif)
+        end
+      end
+
+      it 'Should report error if invalid arguments used for option' do
```
joshua-ostrom, line 134: Nice job taking the time to include spec coverage for this.

lib/salus.rb
ghbren: updated accordingly.

lib/salus/processor.rb
ghbren: Yes, thor supports it.  
I updated my code to work without quotes.

joshua-ostrom:
This is a good candidate to be broken out to it's own utility but this works as a first pass 👍 🙇‍♂️

joshua-ostrom (approved):

joshua-ostrom (approved):

PR #471, opened 2021-11-01
Add --reports Flag to Filter Reports From CLI

geraldnash-cb:


lib/salus/report.rb
```diff
@@ -222,9 +224,26 @@ def publish_report(directive)
       end
     end
 
+    def satisfies_filter?(directive)
+      return true if @report_filter == 'all'
```
ghbren, line 228: The ticket says this would be `'*'`, not `'all'`.  
I'm ok with either. @joshua-ostrom, does it need to be replaced with `'*'`?

Ticket:
```
--reports name:* # run all reports
```

ghbren:
The ticket also says support
```
--reports type:sarif # run all sarif reports
```

This doesn't seem supported in your code.  I'm not seeing any code where you extract the `type` keyword to get the report type.  Did I miss something?

lib/salus/report.rb
```diff
@@ -222,9 +224,26 @@ def publish_report(directive)
...
+      return false if @report_filter == 'none'
+
+      # Use regex to capture two parts of the filter, delimited by a ':'
+      filter_regex = Regexp.new('([a-zA-Z0-9]+):?([a-zA-Z0-9\*]*)')
```
ghbren, line 232: It doesn't seem necessary to use a regex.
I think all you have to do is something like
```ruby
key, value = @report_filter.split(':')
```

spec/lib/salus/report_spec.rb
```diff
@@ -349,4 +349,230 @@ def build_report(report_uri)
...
+      http_url_two = 'https://nerv.tk4/salus-report2'
+      file_path = './spec/fixtures/report/salus_report.json'
+      directives = [
+        { 'uri' => http_url_one, 'format' => 'yaml' },
+        { 'uri' => file_path, 'format' => 'json' },
+        { 'uri' => http_url_two, 'format' => 'yaml' }
+      ]
+      report = build_report(
+        directives,
+        'format:yaml'
```
ghbren, line 549: @joshua-ostrom The ticket you created says
```ruby
--reports type:sarif # run all sarif reports
```
So you were saying the keyword should be `type`, not `format`.
Was this a typo in the ticket?

ghbren:
This PR appears incomplete. The ticket says
```
--reports name:* # run all reports
```
(though I think Josh meant `--reports name:"*".  The `*` without quotes will trigger a shell error)

I don't see any code that extracts the `name` keyword.

ghbren:
Gerald, after you finish updating the PR, please test with running salus directly from the command line to ensure that are no strange errors.
```
docker build -t salus-local .
docker run --rm -t -v $(pwd):/home/repo salus-local --reports ...
```

lib/salus/report.rb
```diff
@@ -222,9 +224,26 @@ def publish_report(directive)
...
+      captured_parameters = filter_regex.match(@report_filter)
+      directive_key = captured_parameters[1]
+      directive_value = directive[directive_key]
+      # Since the value of the second capture group can follow a regex
+      # pattern (i.e. can use wildcards), we build a regex from it
+      # and use it to match against the directive value
+      directive_value_regex = Regexp.new(captured_parameters[2])
+
+      directive.key?(directive_key) && directive_value_regex.match?(directive_value)
```
ghbren, line 241: Not all keys will appear as part of directive.
For example, this is part of the SECT ticket, but `name` will never appear as part of the `report section`.
```
--reports name:* # run all reports
```

Report sections look like below, with no `name` keyword.
```yaml
reports:
  - uri: file://out.json
    format: json
  - uri: file://sarif.json
    format: sarif
```

ghbren: @joshua-ostrom How did you want the code to work with `name`? What are the possible values for `name`?  Only `*`?  

Note the ticket says
```
--reports none # don't run reports
--reports name:* # run all reports
```

Should that instead be
```
--reports none # don't run reports
--reports * # run all reports
```

ghbren:
Gerald, after you finish updating the PR, please test with running salus directly from the command line to ensure that are no strange errors.
```
docker build -t salus-local .
docker run --rm -t -v $(pwd):/home/repo salus-local --reports ...
```

Please test all example commands in the SECT ticket and other commands you could come up with.

Please also make sure your unit tests cover all the asks in the SECT ticket, that includes "none", "name", "type", "uri", if these test cases do not already exist.

spec/lib/salus/report_spec.rb
joshua-ostrom: Good catch, yes format would be great!

lib/salus/report.rb
joshua-ostrom:
```reports:
  - uri: file://out.json
    format: json
    name: arcus
  - uri: file://sarif.json
    format: sarif
    name: vuln_inventory
  - uri: file://sarif2.json
    format: sarif

```

none # filter away all reports, don't create any reports.  
name:* # Run all reports (same as not specifying --reports)
name: arcus # run only the arcus report
name:arcus vuln_inventory # run both of these reports

By using --reports key:value_filter the thought was we could support filtering of multiple fields (for example format and name initially but URI etc in the future).  That said, I'm not tied to that exact format there's latitude if we want to change it up.

* and none are just there for convenience.  We can probably drop * if needed (as the user can just not use --reports CLI in that case).  None would be the same as --reports name:foo (assuming there is no report named foo).

joshua-ostrom:
> @joshua-ostrom , the ticket says
> 
> ```
> --reports name:* # run all reports
> --reports uri:http* # run all http reports
> ```
> 
> did you mean like `--reports name:"*"`? The `*` without quotes will trigger a shell error

See note below on * 👍

lib/salus/report.rb
geraldnash-cb: @joshua-ostrom @ghbren I added an `all` value in place of `--reports name:*` since it mirrors `none`. Using `all` is just like not specifying the flag at all.

lib/salus/report.rb
geraldnash-cb: I went with a regex, because it's a more natural approach to wildcard filter. Also, I can change the regex to support all `-` in addition to other special characters with something like `([a-zA-Z0-9-_\\\.]+):?([a-zA-Z0-9\*]*)`.

ghbren: `-` is just an example, we may allow other characters too.

@joshua-ostrom  Any thoughts on this?  This is just breaking like
`format: abcd` into key `format` and value `abcd`.

lib/salus/report.rb
```diff
@@ -228,17 +228,13 @@ def satisfies_filter?(directive)
...
-      captured_parameters = filter_regex.match(@report_filter)
-      directive_key = captured_parameters[1]
-      directive_value = directive[directive_key]
-      # Since the value of the second capture group can follow a regex
-      # pattern (i.e. can use wildcards), we build a regex from it
-      # and use it to match against the directive value
-      directive_value_regex = Regexp.new(captured_parameters[2])
-
-      directive.key?(directive_key) && directive_value_regex.match?(directive_value)
+      recovered_values = @report_filter.split(':')
```
ghbren, line 231: This `split` probably works for now (assuming there is only one `:`).  But if there is more than one `:` (maybe in the future we'll have something that has more than one `:`), then the code will break.

So it would be safer to use the code I mentioned earlier
```ruby
first_colon = index of the first colon
key = @report_filter[0... first_colon - 1]
value = @report_filter[first_colon + 1 ... end]
```

lib/salus/report.rb
```diff
@@ -228,17 +228,13 @@ def satisfies_filter?(directive)
...
+      filter_key = recovered_values[0]
+      filter_value = recovered_values[1]
+
```
ghbren, line 234: It would be good to check if the key/value are `nil` or empty strings.

lib/salus/report.rb
ghbren: Actually, I think it's already possible to have more than one `:` now, like
```ruby
uri:http://
```

lib/salus/report.rb
```diff
@@ -222,9 +224,22 @@ def publish_report(directive)
...
+    def satisfies_filter?(directive)
+      return true if @report_filter == 'all'
+      return false if @report_filter == 'none'
+
+      recovered_values = @report_filter.split(':')
+      filter_key = recovered_values[0]
+      filter_value = recovered_values[1]
+
+      directive.key?(filter_key) && (
+        directive[filter_key] == filter_value || directive[filter_key] == '*'
```
ghbren, line 236: Shouldn't you check if something starts with `filter_value`, instead of `==`?
if your unit test didn't catch this, then you may want to update your unit tests.

ghbren: If I remember correctly, there was a related bug when you were doing the regex stuff.  The regex match was not looking for lines that starts with a pattern, but was looking for lines that included the pattern.

geraldnash-cb: I actually started with using `start_with?()` but decided to use `==` to prevent the filter from including things we don't want. For example, `sarif` is a prefix of `sarif_diff`  and `http://example.com/dir` is a prefix of `http://example.com/dir/subdir` so I think the added strictness helps.

lib/salus/report.rb
geraldnash-cb: Used the second argument of `split?()` to achieve this.
`'a:b:c:d://:e:'.split(':', 2)` returns `["a", "b:c:d://:e:"]` for example.

lib/salus/report.rb
ghbren: hmm, what if the pattern is `url:http`, then it will not match anything (unless the url really is just `http`, right?

lib/salus/report.rb
```diff
@@ -229,9 +231,28 @@ def publish_report(directive)
...
     end
 
+    def satisfies_filter?(directive)
+      return true if @report_filter == 'all'
+      return false if @report_filter == 'none'
+
+      # Split the filter string only by the first colon
+      recovered_values = @report_filter.split(':', 2)
+      filter_key = recovered_values[0]
+      filter_value = recovered_values[1]
```
ghbren, line 241: `satisfies_filter?` needs to be refactored.
Right now it is invoked once for each uri, but things like `filter_key` and `filter_value` only need to be calculated once.
This also causes an issue if there is a poorly formatted report filter, because if there are 5 report uris, then the code below will print out this error 5 times.
```
Could not send Salus report: (Salus::Report::ExportReportError: Poorly formatted report filter found. Filter key and pattern must be non-empty strings)
```

and maybe you don't need to call `satisfies_filter?` if `@report_filter` is `all` or `none`.

ghbren:
Could you add a short description about --reports here https://github.com/coinbase/salus/blob/master/docs/salus_reports.md with examples?

lib/salus/report.rb
```diff
@@ -229,9 +231,28 @@ def publish_report(directive)
...
+
+      if filter_key.nil? || filter_key == '' || filter_value.nil? || filter_value == ''
```
ghbren, line 243: This code is fine, you don't need to change it. But if you ever want to make code like this shorter, you could use
```ruby
if filter_key.to_s == '' || filter_value.to_s == ''
```

ghbren (approved):

PR #475, opened 2021-11-09
support full sarif diff with --git_diff

ghbren:
Support for `--git_diff` with `--sarif_diff_full`.
The argument to `--git_diff` is a file with the result of `git diff $sha1 $sha2`.  
It filters out sarif diff results that are <B>most likely</B> not included in the git diff.

This PR simply filters out sarif results where each line of the code snippets appears in the git diff as newly added code.  Note we do not check if the code snippets in the sarif results appear as consecutive lines in the newly added code.

Only Bandit, Gosec and PatternSearch are supported.   Other static analysis scanners are not supported because their code snippets do not keep the original intentions or do not show complete lines. This means there is no efficient algorithm to check if the snippets exist in the git diff that are easy to implement (we have time constraints).

Ex.
```
docker run --rm -t -v $(pwd):/home/repo salus-local --sarif_diff_full v2.json v1.json --git_diff v1_diff.txt
```


joshua-ostrom (dismissed):
It would be a win to move the functionality of diffing (both SARIF and git) to a stand alone utility adjacent to salus [same repo is fine].  Approving it would be a nice win to begin pulling this logic into a dedicated script.

joshua-ostrom (approved):

PR #478, opened 2021-11-23
return exit failure if sarif diff file has vulns

ghbren:
return exit failure if sarif diff file has vulns from enforced scanners.

### DIFF
--- a/lib/sarif/base_sarif.rb
+++ b/lib/sarif/base_sarif.rb
@@ -166,6 +166,15 @@ module Sarif
       end
     end
 
+    def self.passed?(sarif)
+      sarif['runs'].each do |run|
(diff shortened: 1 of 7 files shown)


lib/sarif/base_sarif.rb
```diff
@@ -166,6 +166,15 @@ def sarif_level(severity)
       end
     end
 
+    def self.passed?(sarif)
```
joshua-ostrom, line 169: Can we change Sarif::BaseSarif.passed? to Sarif::BaseSarif.salus_passed?

That might better convey the nuance between no SARIF issues found vs issues only found from enforced scanners [our salus logic for pass / fail]

joshua-ostrom (approved):

PR #481, opened 2021-12-02
Recursion Support

joshua-ostrom:
PR adds support for new scanner configuration settings to support recursive scanning:

Example config
```
    recursion: # optional recusion settings.  
      directory_exclusions: # Directories to exclude when recursing.  
      # Any matches that occur in the list of directory_exclusions
      # will be ignored
        - vendor
      directories: # Directories to recurse into.  When present,
      # any directories that need to be scaned should explicity listed
        - ./
        - payments/lhv
        - infra/sso/identity_provider
      directories_matching: # Dynamically identify directories to recurse into.
      # Each entry may have a combination of filename and or content.  Filename
      # matches on the name of the file, while content matches on the content within
      # the file.
        - 
          filename: "BUILD.bazel"
          content: "bundle//:rails"
        - 
          filename: "package.json"
      static_files: # Files to copy from root directory to recused directories
      # This can be useful for mono repositories where a common file should be 
      # copied to sub directories for proper scanning.
        - Gemfile
        - Gemfile.lock
```
The allows Salus to scan more then just the project's root directory.  Users can configure  explicit directories to scan or they can dynamically determine what directories should be scanned by a given scanner.  Exclusions (block lists) are provided as well as the ability to copy static files to recused directories if needed by a given scanner.

SARIF has been updated to capture the SRCROOT accordingly.  

Scanning is still sequential with concurrent scanning to be addressed in a following PR.


geraldnash-cb (approved):

cb-heimdall:
Review Error for geraldnash-cb @ 2021-12-02 15:14:46 UTC

geraldnash-cb (approved):

geraldnash-cb (approved):

ghbren:
Should you check whether certain operations may be reading/writing files to places where they don't have permission for?    There is an existing `safe_local_report_path?` in `report.rb`.

ghbren:
Have you tested with?
```
docker build ... local_salus
docker run  ... local_salus ...
```

joshua-ostrom:
> docker run  ... local_salus ...



> Should you check whether certain operations may be reading/writing files to places where they don't have permission for? There is an existing `safe_local_report_path?` in `report.rb`.

Ah great call out yes, can update to use that helper

joshua-ostrom:
> Have you tested with?
> 
> ```
> docker build ... local_salus
> docker run  ... local_salus ...
> ```

Yes that's how I generated the SARIF to test salus bot w/mono repo findings

lib/salus/repo_searcher.rb
```diff
@@ -0,0 +1,201 @@
...
+    end
+
+    ##
+    # @param [Array] args CLI rg command and options to run
+    # @return [Array<String>] Relative path of files returned
+    # from running the command.
+    def run_rg(*args)
+      data = nil
+      Dir.chdir(@path_to_repo) do
+        data = IO.popen(args).read
```
ghbren, line 133: What if `rg` throws exception or outputs something unexpected?

joshua-ostrom: We will fail - open to thoughts, if we are unable to scan the expected / configured directories I'm okay with a fail.

ghbren (approved):

PR #486, opened 2021-12-08
ruby license scanner, first commit

Ojotolu:


lib/salus/scanners/report_ruby_gems.rb
```diff
@@ -65,19 +71,92 @@ def record_dependencies_from_gemfile
...
+      dice
+    end
+
+    # Converts each license in an array of licenses to spdx formatted licenses
+    def to_spdx(licenses_arr)
+      licenses_arr.map{ |license| match_license(license)}
+    end 
+
+    # Compares reported license with spdx licenses and returns closest match 
+    def match_license(license)
```
ghbren, line 129: The actual license type in spdx format for each license should be cached.  
There is no need to re-process each license  if the spdx license is in the cache.

lib/salus/scanners/report_ruby_gems.rb
```diff
@@ -65,19 +71,92 @@ def record_dependencies_from_gemfile
...
     def record_ruby_gem(name:, version:, source:, dependency_file:)
-      report_dependency(
+      dep_hsh = report_dependency(
         dependency_file,
         type: 'gem',
         name: name,
         version: version,
-        source: source
+        source: source,
+        licenses: @opt_hsh[[name,version].to_s]
```
ghbren, line 155: If the report ruby license part throws an exception, then`@opt_hsh[[name,version]` may evaluate to `nil`.  
But then if we have `licenses: nil`, then there is a good chance this will break other things.

I think the fix is to not add `licenses: ...` to `dep_hash` if `[name,version]` is `nil`.
You should also add a test case for this.

Gemfile
```diff
@@ -13,23 +13,24 @@ gem 'rgl', '~> 0.5.7'
 gem 'safe_yaml', '~> 1.0'
 gem 'thor', '~> 1.1.0'
 gem 'toml', '~> 0.2.0'
+gem 'license_finder'
 
 group :scanners do
   gem 'brakeman', '4.10.0'
   gem 'bundler-audit', '~> 0.8.0'
 end
```
joshua-ostrom, line 22: be sure to restore the original groups

lib/cyclonedx/base.rb
```diff
@@ -21,6 +21,7 @@ def build_components_object
       info[:dependencies].each do |dependency|
         components << parse_dependency(dependency)
       end
+      puts "this is components #{components}"
```
joshua-ostrom, line 24: Be sure to axe the various puts statements

lib/cyclonedx/base.rb
```diff
@@ -43,10 +44,16 @@ def build_component(dependency)
         "group": "", # TODO: add group or domain name of the publisher
         "name": dependency[:name],
         "version": version_string(dependency),
-        "purl": package_url(dependency)
+        "purl": package_url(dependency),
+        "licenses": get_license(dependency)
       }
     end
 
+    # takes an array of spdx formatted licenses and returns a hash conforming to CycloneDX format
```
joshua-ostrom, line 52: 👍 for the description

lib/cyclonedx/report.rb
```diff
@@ -58,10 +58,11 @@ def to_cyclonedx
 
     def self.validate_cyclonedx(cyclonedx_report)
       cyclonedx_string = JSON.pretty_generate(cyclonedx_report)
+      puts "\n\n\nthis is a test 8888388231823 ####\n\n\n"
```
joshua-ostrom, line 61: ditto on puts

lib/salus/report.rb
```diff
@@ -195,6 +195,7 @@ def to_cyclonedx(config = {})
         projectVersion: "1",
         bom: Base64.strict_encode64(JSON.generate(cyclonedx_bom))
       }
+      File.write("tolu_test.json", JSON.generate(cyclonedx_bom))
```
joshua-ostrom, line 198: be sure to drop this

Gemfile
ghbren: please add the version for license finder, like the other gems

joshua-ostrom:
Be sure to fill out the PR description and hit the failing specs

lib/salus/scanners/report_ruby_gems.rb
```diff
@@ -1,13 +1,18 @@
 require 'salus/scanners/base'
+require 'salus/dice_coefficient'
+
```
ghbren, line 3: check on dice coefficient

lib/salus/scanners/report_ruby_gems.rb
```diff
@@ -65,19 +71,89 @@ def record_dependencies_from_gemfile
...
         )
       end
     end
 
+    # Converts a JSON string to hash and returns an array of license information
+    def find_licenses 
+      output = run_license_finder()
+      license_finder_parsed = output.split('approval:').last
+      license_finder_hsh = JSON.parse(license_finder_parsed)
```
ghbren, line 84: the output prints out something like 
```
.... approval:
 {license_info_hash}
```

ghbren:
replaced by https://github.com/coinbase/salus/pull/511

PR #492, opened 2021-12-18
add line of code for bundle audit CVEs

ghbren:
Add line of code for bundle/npm/yarn audit CVEs.

Tested with
* new unit specs
* docker run local salus and manually examined txt, json and sarif outputs


lib/salus/scanners/bundle_audit.rb
```diff
@@ -44,6 +45,18 @@ def run_scanner(scanner, ignore)
         hash = serialize_vuln(result)
         @vulns.push(hash)
 
+        pattern = case hash[:type]
+                  when 'UnpatchedGem'
+                    hash[:name] + ' (' + hash[:version] + ')'
+                  when 'InsecureSource'
+                    hash[:source]
+                  end
+        if !pattern.nil?
```
joshua-ostrom, line 54: Minor, can you move the line_number logic to a stand alone method

lib/salus/scanners/yarn_audit.rb
```diff
@@ -46,6 +46,7 @@ def run
         # vulns were all whitelisted
         return report_success if vulns.empty?
 
+        add_line_number(vulns)
```
joshua-ostrom, line 49: Another minor but I'd be a fan of moving the logic you've added here to a stand alone class for working with the line number pieces.  We will have a coupling of passing the vulns hash around but if we document the required hash keys I think it'll be a nice separation of logic here.
```
YarnLock.new('yarn.lock').add_line_number(vulns)
```

lib/salus/scanners/bundle_audit.rb
ghbren: done

lib/salus/scanners/yarn_audit.rb
ghbren: done

lib/salus/package_utils.rb
```diff
@@ -0,0 +1,6 @@
+Dir.entries(File.expand_path('package_utils', __dir__)).sort.each do |filename|
```
joshua-ostrom, line 1: :+1 for the sort... that race condition still haunts me :-)

lib/salus/package_utils/gemfile_lock.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    def add_line_number(cve)
+      pattern = case cve[:type]
+                when 'UnpatchedGem'
+                  cve[:name] + ' (' + cve[:version] + ')'
+                when 'InsecureSource'
+                  cve[:source]
+                end
+      if !pattern.nil?
+        line_in_gemfile_lock = `grep -n " #{pattern}" #{@path}`
```
joshua-ostrom, line 15: `IO.popen(["grep", "-n", pattern, @path]) ` 
would offer some protection against any malicious content... that's a small risk here but I think if someone hijacked the vuln database and injected a malicious string as the "version" we could pick it up here.   Probably paranoid to worry about it but I'm becoming a fan of IO.popen over backticks. 

Thoughts?  If I'm being too paranoid ignore me!

joshua-ostrom:
One small security callout, see what you think... I think your test coverage will suffice since we refactored here it's more of an end-end test of the new class than unit coverage but it looks like the behavior is tested.  

Thanks Brianna!

joshua-ostrom (approved):

PR #507, opened 2022-01-04
Prevent NPMAudit From Adding Allowlisted Vulns to the SARIF Results List

geraldnash-cb:


lib/sarif/npm_audit_sarif.rb
```diff
@@ -20,6 +20,7 @@ def parse_scan_report!
     def parse_issue(issue)
       id = issue[:id].to_s
       return nil if @issues.include?(id)
+      return nil if @exceptions.include?(id)
```
ghbren, line 24: Minor nit. The two lines above can be combined to something like
```ruby
return nil if @issues.include?(id) || @exceptions.include?(id)
```

Or
```ruby
return nil if [@issues, @exceptions].any? { |i| i.include?(id) }
```

ghbren (approved):

PR #511, opened 2022-01-06
Add ruby license scanner

saikat056:
This PR is the continuation of this [work](https://github.com/coinbase/salus/pull/486). In the PR, we gather the spdx license information of ruby gems and incorporate them to CycloneDx report. In order to match proper spdx formatted license for a particular license, we have used [dice coefficient](https://en.wikibooks.org/wiki/Algorithm_Implementation/Strings/Dice%27s_coefficient)

Following is a screenshot of how the `bom` looks like with license information:

Screen Shot 2022-01-14 at 10 26 46 AM

We are using https://guides.rubygems.org/rubygems-org-api-v2/ api to find information about ruby gem licenses. Ruby gems can be downloaded from multiple sources. But, here we are only targeting rubygems.org.

### DIFF
--- a/lib/cyclonedx/report_ruby_gems_cyclonedx.rb
+++ b/lib/cyclonedx/report_ruby_gems_cyclonedx.rb
@@ -6,5 +6,17 @@ module Cyclonedx
     def initialize(scan_report, config = {})
       super(scan_report, config)
     end
+
+    def build_component(dependency)
+      super.merge({
+                    "licenses": licenses_for(dependency)
+                  })
+    end
+
+    def licenses_for(dependency)
+      return [] if dependency[:licenses].nil?
+
+      dependency[:licenses].map { |license| { "license" => { "id" => license } } }
+    end
   end
 end
--- a/lib/salus/dice_coefficient.rb
+++ b/lib/salus/dice_coefficient.rb
@@ -1,0 +1,23 @@
+module Salus
+  class DiceCoefficient
+    # source: https://en.wikibooks.org/wiki/Algorithm_Implementation/Strings/Dice%27s_coefficient
+    # Dice co-efficient can be used for measuring the similarity of two strings
+    # bigram: A bigram or digram is a sequence of two adjacent elements
+    #         from a string of tokens, which are typically letters,
+    #         syllables, or words.
+    #         For example, bigram of a string ('hello') can be written as the following requence
+    #         [["h", "e"], ["e", "l"], ["l", "l"], ["l", "o"]]
+    #
+    # dice coefficient = bigram overlap * 2 / (bigrams in a + bigrams in b)
+    def self.dice(string_a, string_b)
+      a_bigrams = string_a.each_char.each_cons(2).to_a
+      b_bigrams = string_b.each_char.each_cons(2).to_a
+
+      overlap = (a_bigrams & b_bigrams).size
+
+      total = a_bigrams.size + b_bigrams.size
+
+      overlap * 2.0 / total
+    end
+  end
+end
--- a/lib/salus/scanners/report_ruby_gems.rb
+++ b/lib/salus/scanners/report_ruby_gems.rb
@@ -1,9 +1,17 @@
 require 'salus/scanners/base'
+require 'salus/dice_coefficient'
+require 'uri'
+require 'net/http'
 
 # Report the use of any Ruby gems.
-
 module Salus::Scanners
   class ReportRubyGems < Base
+    class RubyGemsApiError < StandardError; end
+    class ApiTooManyRequestsError < StandardError; end
+
+    SPDX_SCHEMA_FILE = 'lib/cyclonedx/schema/spdx.schema.json'.freeze
+    MAX_RETRIES_FOR_RUBY_GEMS_API = 2
+
     def run
       # A lockfile is the most definitive source of truth for what will run
       # in production. It also lists the dependencies of dependencies.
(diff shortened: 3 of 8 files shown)


saikat056:
@ghbren It would be great if you please review this PR. Thank you.

lib/salus/dice_coefficient.rb
```diff
@@ -0,0 +1,23 @@
...
+  class DiceCoefficient
+    # source: https://en.wikibooks.org/wiki/Algorithm_Implementation/Strings/Dice%27s_coefficient
+    # Dice co-efficient can be used for measuring the similarity of two strings
+    # bigram: A bigram or digram is a sequence of two adjacent elements
+    #         from a string of tokens, which are typically letters,
+    #         syllables, or words.
+    #         For example, bigram of a string ('hello') can be written as the following requence
+    #         [["h", "e"], ["e", "l"], ["l", "l"], ["l", "o"]]
+    #
+    # dice coefficient = bigram overlap * 2 / (bigrams in a + bigrams in b)
```
ghbren, line 11: 💯   for the documentation

lib/salus/scanners/report_ruby_gems.rb
```diff
@@ -76,8 +80,50 @@ def record_ruby_gem(name:, version:, source:, dependency_file:)
...
-        source: source
+        source: source,
+        licenses: find_licenses_for(name, version)
       )
     end
+
+    def find_licenses_for(gem_name, version)
+      res = Net::HTTP.get_response(ruby_gems_uri_for(gem_name, version))
+
+      return [] unless res.is_a?(Net::HTTPSuccess)
```
ghbren, line 91: If `res` is not success, then a warning should be logged.

ghbren:
```ruby
 report_warn(:gem_url_bad_resp, "bad response for #{gem_name} #{version}")
```

lib/salus/scanners/report_ruby_gems.rb
```diff
@@ -76,8 +80,50 @@ def record_ruby_gem(name:, version:, source:, dependency_file:)
...
+      msg = "Unable to gather license information " \
+        "using rubygems api" \
+        "with error message #{e.class}: #{e.message}"
+      bugsnag_notify(msg)
+
+      []
+    end
+
+    def ruby_gems_uri_for(gem_name, version)
+      URI("https://rubygems.org/api/v2/rubygems/#{gem_name}/versions/#{version}.json")
```
ghbren, line 105: I wonder how common this would cause conflicts, because not all gems are pulled from rubygems.org. 

Some issues:
* Gem A from rubygems.org and random_server.com may have difference license info
* Gem A exists on random_server.com, but not rubygems.org
  * If this is too common, then we will get a large number of bugsnag messages and pager duty may ping us crazy.

lib/salus/scanners/report_ruby_gems.rb
```diff
@@ -76,8 +80,50 @@ def record_ruby_gem(name:, version:, source:, dependency_file:)
...
       )
     end
+
+    def find_licenses_for(gem_name, version)
+      res = Net::HTTP.get_response(ruby_gems_uri_for(gem_name, version))
+
+      return [] unless res.is_a?(Net::HTTPSuccess)
+
+      licenses = JSON.parse(res.body)['licenses']
+      licenses.map { |license| spdx_license_for(license) }
```
ghbren, line 94: Are licenses from the json already in spdx format?
If not, do you know roughly what percentage of dice coefficient conversions are accurate?

lib/salus/scanners/report_ruby_gems.rb
ghbren: If we decide to not fix this, then we should clearly communicate this and other issues with the scanner in the
* PR description 
* new salus version release notes.

spec/lib/salus/scanners/report_ruby_gems_spec.rb
```diff
@@ -2,6 +2,10 @@
 
 describe Salus::Scanners::ReportRubyGems do
```
ghbren, line 3: It would be helpful to also check that the end json can be validated against the cyclonedx schema (we did this for sarif).  
We tried briefly to find an existing function to do this, but couldn't find one.
If this is time consuming, then ignore the comment.

lib/cyclonedx/report_ruby_gems_cyclonedx.rb
```diff
@@ -6,5 +6,17 @@ class ReportRubyGems < Base
     def initialize(scan_report, config = {})
       super(scan_report, config)
     end
+
+    def build_component(dependency)
+      super.merge({
```
ghbren, line 11: Probably a stupid question. But what does `super.merge` do here?
I don't see `merge` defined in the super class.

lib/salus/scanners/report_ruby_gems.rb
saikat056: We are focusing license information about external ruby gems which come from rubygems.org. The internal ruby gems come from another source. That's why it would be great if we don't send message to bugsnag for that.

lib/salus/scanners/report_ruby_gems.rb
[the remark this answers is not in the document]
saikat056: Good question. We will try to avoid that because of the reason described [here](https://github.com/coinbase/salus/pull/511#discussion_r785164359).

lib/salus/scanners/report_ruby_gems.rb
saikat056: https://guides.rubygems.org/rubygems-org-api-v2/ does not mention that it's strictly following spdx format. In that case, it is better to use dice coefficient to match the license.

According to [this](https://towardsdatascience.com/metrics-to-evaluate-your-semantic-segmentation-model-6bcb99639aa2), dice coefficient is highly used for segmentation matching. I have not found any article/writing about mathematical analysis on dice coefficient. Another name of dice coefficient is [F-score](https://en.wikipedia.org/wiki/F-score). We can talk about them if you want.

lib/cyclonedx/report_ruby_gems_cyclonedx.rb
saikat056: Good question. This hash [merge](https://apidock.com/ruby/Hash/merge) of ruby. The `build_component` method of superclass returns a hash. In the Cyclonedx::ReportRubyGems, we are merging license information. This is done this way, because we will have license information available to Ruby gems for now.

spec/lib/salus/scanners/report_ruby_gems_spec.rb
saikat056: I will look into this.

lib/salus/scanners/report_ruby_gems.rb
saikat056: Updated this part. Thanks.

lib/salus/scanners/report_ruby_gems.rb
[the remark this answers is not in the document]
saikat056: Updated this part.

ghbren (approved):

geraldnash-cb (approved):

PR #512, opened 2022-01-06
Concurrent Scanning

joshua-ostrom:
# Concurrent Scanning

This update changes our scanning from sequential to concurrent.  

References to the thread unsafe Dir.chdir were refactored.  Bundle Audit likewise uses this and required a patch.  We aim to create a PR against that project as a proper long term solution.

Tested against 6+ production repos, covering both recursive and non-recursive scans


lib/salus/scanners/semgrep.rb
```diff
@@ -127,51 +95,82 @@ def run
...
+                required: match["required"],
+                msg: msg,
+                hit: hit_to_string(hit, base_path)
               }
             end
-          else
+          end
+
+        # possible exit codes from https://github.com/returntocorp/semgrep/blob/9ac58092cb8ac02bb1f41f59808d4f03a5b8206e/semgrep/semgrep/util.py#L11-L18
+        elsif [1, 2, 3, 4, 5, 6, 7].include?(shell_return.status)
```
saikat056, line 118: Following can be used as short-cut:

(1..7).to_a.include?(3)

ghbren (approved):

lib/salus/scanners/semgrep.rb
joshua-ostrom: Thanks Saikat, I'll make the update, agree it's clearer. 👍 

This was existing code that git diff is picking up from unwinding the Dir.chdir(@repository.path_to_repo) do block.

lib/salus/processor.rb
```diff
@@ -93,42 +95,84 @@ def fetch_config_file(source_uri, repo_path)
 
     def scan_project
```
saikat056, line 96: method 'def scan_project' has become long. Is it possible to extract some part of it to a separate method?

lib/salus/processor.rb
```diff
@@ -103,14 +103,23 @@ def scan_project
...
+          end
+
+          # We won't auto cleanup - we'll manually delete any copied fields after
+          # the various threads have finished
+          copied = RepoSearcher.new(@repo_path, config, false).matching_repos do |repo|
             threads << Thread.new do
               scanner = scanner_class.new(repository: repo, config: config)
 
-              unless @config.scanner_active?(scanner_name) && scanner.should_run?
+              unless scanner.should_run?
```
ghbren, line 122: Why was `@config.scanner_active?(scanner_name)` removed from the condition?

ghbren (approved):

lib/salus/scanners/base.rb
```diff
@@ -112,22 +117,27 @@ def run!(salus_report:, required:, pass_on_raise:, reraise:)
...
+                  chdir: File.expand_path(@repository&.path_to_repo))
       # If we're passed a string, convert it to an array before passing to capture3
       command = command.split unless command.is_a?(Array)
-      Salus::PluginManager.send_event(:run_shell, command)
-      Salus::ShellResult.new(*Open3.capture3(env, *command, stdin_data: stdin_data))
+      Salus::PluginManager.send_event(:run_shell, command, chdir: chdir)
+      #  chdir: '/some/directory'
+      opts = { stdin_data: stdin_data }
+      opts[:chdir] = chdir unless chdir.nil? || chdir == "."
+      Salus::ShellResult.new(*Open3.capture3(env, *command, opts))
```
saikat056, line 140: Non-blocking question:
Is `Open3.capture3` with `opts[:chdir]` thread-safe? We know that `Dir.chdir `is not thread-safe. 🤔

saikat056:
Looks good to me. Approved. 👍

lib/salus/scanners/base.rb
joshua-ostrom: Yes, a good discussion here https://bugs.ruby-lang.org/issues/9785.  The new process spawned is uses chdir vs attempting to change the ruby process itself.

ghbren (approved):

PR #519, opened 2022-01-25
Add Maven Project Dependency Reporter

geraldnash-cb:


bin/parse_pom_xml
```diff
@@ -0,0 +1,52 @@
...
+# Write a JSON string of the dependencies of a Java Maven project
+# to stdout using its pom.xml project file
+
+require 'json'
+
+dependencies_regex = %r{
+    <dependencies>
+        [\s\S]*
+    <\/dependencies>
+}x
```
ghbren, line 12: What do `%r{` and `}x` mean?   Is it a way to enclose a multiline regex?

bin/parse_pom_xml
```diff
@@ -0,0 +1,52 @@
...
+
+dependency_metadata_regex = %r{
```
ghbren, line 14: This regex assumes the keys follow a certain order, but the actual order could be random.
In addition, it assumes the file follows a specific format (with the newlines, spaces, etc.) but files that do not follow this format could be valid too.

ghbren: There is likely existing code that parses xmls.  Could you google it?

bin/parse_pom_xml
```diff
@@ -0,0 +1,52 @@
...
+        (<scope>
+            (?<scope>\S+)
+        <\/scope>\s*)?
+        (<type>
+            (?<type>\S+)
+        <\/type>\s*)?
+    <\/dependency>
+}x
+
+dependency_lists = File.read(ARGV[0]).scan(dependencies_regex)
```
ghbren, line 34: `File.read` should be checked for file not found exception.

bin/parse_pom_xml
```diff
@@ -0,0 +1,52 @@
...
+  # Parse individual <dependency> entries
+  dependency_list.scan(dependency_metadata_regex).each do |dependency_properties|
+    dependency_hash = {}
+    dependency_hash[:group_id] = dependency_properties[0]
+    dependency_hash[:artifact_id] = dependency_properties[1]
+    dependency_hash[:version] = dependency_properties[2]
+    dependency_hash[:scope] = dependency_properties[3]
+    dependency_hash[:type] = dependency_properties[4]
+    dependencies.append(dependency_hash)
+  end
```
ghbren, line 49: I did not review the `dependency_lists.each` block because you will likely modify this part to work with an existing xml parser.

lib/salus/scanners/report_pom_xml.rb
```diff
@@ -0,0 +1,38 @@
...
+
+# Report Java library usage
+
+module Salus::Scanners
+  class ReportPomXml < Base
+    def run
+      shell_return = run_shell("bin/parse_pom_xml #{@repository.path_to_repo}/pom.xml", chdir: nil)
+
+      if !shell_return.success?
+        report_error(shell_return.stderr)
```
ghbren, line 11: Could you call `report_stderr` here too?
I forgot the exact reason that both `report_error` and `report_stderr` are needed, but that's what other scanners do.  I think one of the prints to the screen or something.

lib/salus/scanners/report_pom_xml.rb
```diff
@@ -0,0 +1,38 @@
...
+        return
+      end
+
+      dependencies = JSON.parse(shell_return.stdout)
```
ghbren, line 15: It would be safer to check for could not parse json error, in case weird things somehow got appended to stdout.

lib/salus/scanners/report_pom_xml.rb
ghbren: You may want to check on how `shell_return.success?` is defined. It might look for a specific status code or if you explicitly wrote error to stderr or something.

lib/salus/scanners/report_pom_xml.rb
```diff
@@ -0,0 +1,38 @@
...
+
+      dependencies.each do |dependency|
+        group_id = dependency['group_id']
+        artifact_id = dependency['artifact_id']
+        version = dependency['version']
+        report_dependency(
+          'pom.xml',
+          type: 'maven',
+          name: "#{group_id}.#{artifact_id}",
+          version: version.nil? ? 'unknown' : version
```
ghbren, line 25: I do not know if `unknown` is a valid version value in cyclonedx.  
Could you find a cyclondex schema or something to check?  If it's an invalid value, then when our service processes the the cyclondex, then everything could have blank values (happened with our license scanner).

bin/parse_pom_xml
ghbren: Is it possible for a pom to not have some of the keys in the regex below and still be valid?

ghbren:
Additional test cases should be added for cases where pom could not be parsed, or could be parsed but certain expected keys are missing

bin/parse_pom_xml
geraldnash-cb: Yes, that's correct.

bin/parse_pom_xml
geraldnash-cb: It's my understanding that the only required keys are `groupId` and `artifactId`.

spec/fixtures/report_pom_xml/bad_pom_cant_parse/pom.xml
```diff
@@ -0,0 +1,12 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>connect-api</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+</project>
```
saikat056, line 12: **Non-blocking questions**: Can we put a new line at the end to this warning from GitHub? Can we do the same for other places in this PR?

lib/salus/scanners/report_pom_xml.rb
```diff
@@ -0,0 +1,47 @@
...
+        group_id = dependency['group_id']
+        artifact_id = dependency['artifact_id']
+        version = dependency['version']
+        report_error('No group ID found for a dependency!') if group_id.nil?
+        report_error('No artifact ID found for a dependency!') if artifact_id.nil?
+        report_dependency(
+          'pom.xml',
+          type: 'maven',
+          name: "#{group_id}/#{artifact_id}",
+          version: version.nil? ? 'unknown' : version
```
saikat056, line 34: We don't need to create the variable 'version' if it is used only once in the code.

Can we use a constant UNKNOWN_VERSION instead?

`version: version.nil? ? UNKNOWN_VERSION : dependency['version']`

spec/lib/salus/scanners/report_pom_xml_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+    scanner.run
+
+    dependencies = scanner.report.to_h.fetch(:info).fetch(:dependencies)
+
+    expect(dependencies).to match_array(
+      [
+        {
+          dependency_file: 'pom.xml',
+            name: 'org.apache.kafka/connect-api',
+            version: 'unknown',
```
saikat056, line 50: version: "#{described_class::UNKNOWN_VERSION}" if we have a constant defined according to https://github.com/coinbase/salus/pull/519/files#r793887410

Gemfile
```diff
@@ -8,6 +8,7 @@ gem 'bugsnag', '~> 6.19.0'
 gem 'bundler', '= 2.3.1'
 gem 'faraday', '~> 1.3'
 gem 'github-linguist', '~> 7.13.0'
+gem 'nokogiri', '1.13.1'
```
saikat056, line 11: `gem 'nokogiri', '~> 1.13.1'`

Can we use [pessimistic operator](https://thoughtbot.com/blog/rubys-pessimistic-operator), ~> here so that `bundle install` is able to take the patch releases? Patch releases contain security fixes sometimes?

saikat056:
This change looks good to me. I added some minor suggestions. Good work. 🎉

saikat056:
@geraldnash-cb Can you please rebase this PR? Looks like the branch is out-of-date.

saikat056 (approved):

saikat056 (approved):

lib/salus/scanners/report_pom_xml.rb
```diff
@@ -0,0 +1,48 @@
...
+      dependencies = nil
+      begin
+        dependencies = JSON.parse(shell_return.stdout)
+      rescue JSON::ParserError
+        err_msg = "Could not parse JSON returned by bin/parse_pom_xml's stdout!"
+        report_stderr(err_msg)
+        report_error(err_msg)
+      end
+
+      dependencies.each do |dependency|
```
ghbren, line 26: `dependencies` is set to `nil` above.  It there is an exception, then it will stay `nil`, and then the `dependencies.each` will throw an exception.

ghbren: So I think you should `return` if `JSON.parse` fails.

lib/salus/scanners/report_pom_xml.rb
```diff
@@ -0,0 +1,48 @@
...
+        group_id = dependency['group_id']
+        artifact_id = dependency['artifact_id']
+        report_error('No group ID found for a dependency!') if group_id.nil?
+        report_error('No artifact ID found for a dependency!') if artifact_id.nil?
+        report_dependency(
+          'pom.xml',
+          type: 'maven',
+          name: "#{group_id}/#{artifact_id}",
```
ghbren, line 34: Minor nit: 
If `artifact_id` is nil, then don't add the `/`.

PR #521, opened 2022-01-27
Add sort option for different output formats

Kamsiy:
This PR adds a sort option for the different salus report formats (YAML, json, cyclondx, sarif). This can be added as an optional configuration for reports. For example:

`reports:`
&nbsp;`- uri: file://salus-report2.json`
 &nbsp; &nbsp;`format: sarif`
 &nbsp; &nbsp; `sort: true`


lib/salus/report.rb
```diff
@@ -157,19 +158,27 @@ def to_s(verbose: false, wrap: WRAP, use_colors: false)
       output
     end
 
-    def to_yaml
-      YAML.dump(to_h)
+    def to_yaml(sort = false)
+      sort ? YAML.dump(to_h.deep_sort) : YAML.dump(to_h)
```
ghbren, line 162: Minor nit: You could make the code cleaner (only call dump once) with 
```ruby
res = sort ?  to_h.deep_sort, to_h
YAML.dump(res)
```

Similar for the other methods.

lib/salus/report.rb
```diff
@@ -157,19 +158,27 @@ def to_s(verbose: false, wrap: WRAP, use_colors: false)
       output
     end
 
-    def to_yaml
-      YAML.dump(to_h)
+    def to_yaml(sort = false)
```
ghbren, line 161: I think default should be `true`.

PR #527, opened 2022-01-28
Add Gradle Dependency Reporter

geraldnash-cb:


bin/parse_build_gradle
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env ruby
+
+require 'json'
+
+dependencies_regex = /dependencies\s+{[\s\S]*}/
+
+dependency_metadata_regex =
+  /((implementation)|(classpath))\s['|"](?<classpath>.*):(?<package>.*):(?<version>.+)['|"]/
```
ghbren, line 8: This regex doesn't cover everything, here's an example
`        classpath("com.android.tools.build:gradle:4.1.0")`

bin/parse_gradle_deps
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env ruby
+
+require 'json'
+
+dependency_metadata_regex = /-\s(?<group_id>.+):(?<artifact_id>.+):(?<version>.+)/
```
ghbren, line 5: Is it ever possible for the keys to be in a different order?  In that case, the regex won't work.

Does gradle provide an option to output to a structured format like json?

lib/salus/scanners/report_gradle_deps.rb
```diff
@@ -0,0 +1,46 @@
+require 'salus/scanners/base'
+
+# Report Java library usage
+
+module Salus::Scanners
+  class ReportGradleDeps < Base
+    UNKNOWN_VERSION = ''.freeze
```
ghbren, line 7: It seems like multiple classes use `UNKNOWN_VERSION`. So it would be better to move `UNKNOWN_VERSION` to the Base class.

lib/salus/scanners/report_gradle_deps.rb
```diff
@@ -0,0 +1,46 @@
...
+        return
+      end
+
+      dependencies = nil
+      begin
+        dependencies = JSON.parse(shell_return.stdout)
+      rescue JSON::ParserError
+        err_msg = "Could not parse JSON returned by /home/bin/parse_gradle_deps's stdout!"
+        report_stderr(err_msg)
+        report_error(err_msg)
```
ghbren, line 23: You should `return` here.  
Otherwise, `dependencies` will stay `nil`, and then when `dependencies.each` runs, there will be a `each not defined on nil exception`.

ghbren, line 23: Actually the `dependencies = nil` above can be eliminated.

bin/parse_gradle_deps
geraldnash-cb: Not without modifying the project file itself and using it as a task runner (see this [SO](https://stackoverflow.com/questions/36536561/is-there-a-way-to-list-all-gradle-dependencies-programmatically) page).

ghbren: Is the lock file itself always a yaml? It looks like so from the lockfiles you added to specs.  If so, you can just call yaml parse to get a structured objected of the values.

ghbren: Could you post an example of gradle_stdout?

geraldnash-cb: From the link shared above:
```
+--- org.javassist:javassist:3.13.0-GA -> 3.20.0-GA
+--- org.apache.geronimo.specs:geronimo-jta_1.1_spec:1.1.1
+--- org.hibernate:hibernate-core:5.1.0.Final
|    +--- org.jboss.logging:jboss-logging:3.3.0.Final
|    +--- org.hibernate.javax.persistence:hibernate-jpa-2.1-api:1.0.0.Final
|    +--- org.javassist:javassist:3.20.0-GA
|    +--- antlr:antlr:2.7.7
|    +--- org.apache.geronimo.specs:geronimo-jta_1.1_spec:1.1.1
|    +--- org.jboss:jandex:2.0.0.Final
|    +--- com.fasterxml:classmate:1.3.0
|    +--- dom4j:dom4j:1.6.1
|    |    \--- xml-apis:xml-apis:1.0.b2
|    \--- org.hibernate.common:hibernate-commons-annotations:5.0.1.Final
|         \--- org.jboss.logging:jboss-logging:3.3.0.Final
\--- org.springframework:spring-web:4.2.5.RELEASE
```

lib/cyclonedx/report_gradle_deps_cyclonedx.rb
```diff
@@ -0,0 +1,10 @@
+module Cyclonedx
+  class ReportGradleDeps < Base
+    TYPE = "gradle".freeze
+    UNLOCKED_DEPENDENCY_FILE = "build.gradle".freeze
```
ghbren, line 4: Seems like the two constants aboved  are not used anywhere?

ghbren (approved):

PR #533, opened 2022-02-01
Add CocoaPods (Podfile.lock) Dependency Reporter

geraldnash-cb:


lib/salus/scanners/report_podfile_lock.rb
```diff
@@ -0,0 +1,46 @@
...
+      dependencies = nil
+      begin
+        dependencies = JSON.parse(shell_return.stdout)
+      rescue JSON::ParserError
+        err_msg = "Could not parse JSON returned by bin/parse_podfile_lock's stdout!"
+        report_stderr(err_msg)
+        report_error(err_msg)
+        return
+      end
+
```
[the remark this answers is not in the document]
geraldnash-cb: LGTM! Thanks for the feedback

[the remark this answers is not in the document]
geraldnash-cb: I'm unable to force `dependencies` to be `nil` to directly test the above, but I've also added a test case (that tests `parse_podfile_lock`) for a lock file that simply has a literal `null` value.

bin/parse_podfile_lock
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env ruby
+
+require 'json'
+
+dependency_metadata_regex = /-\s(?<pod>.*?)\s\((=|(~>))?\s?(?<version>(\d*\.?)+)\)/
```
ghbren, line 5: I'm worried about this regex.  Does cocoa provide dependencies in structured formats like json?

geraldnash-cb: Not that I know of. There's a `pod list` command that's supposed to list pods but I'm not able to get it to work locally with `Podfile.lock`s. There was also an individually maintained tool called [`cocoapods-dependencies`](https://github.com/segiddins/cocoapods-dependencies), but it's gone three years without an update now, so I'm not sure we should use it. I'm open to thoughts, though!

ghbren: Is the lock file itself always a yaml? It looks like so from the lockfiles you added to specs.  If so, you can just call yaml parse to get a structured objected of the values.

ghbren: Actually, please take a look at the `to_yaml` method here
https://www.rubydoc.info/github/CocoaPods/Core/Pod/Lockfile#to_yaml-instance_method

geraldnash-cb: Thanks for pointing that out, I didn't notice it's valid YAML. It's now using YAML to parse the dependency list, but because the version of each dependency is in the same string as the name it's still using a (smaller) regex.

Difference is before the regex grabbed every dependency and version of the file. Now the file is parsed as YAML and regex is used only to extract the name and version from each dependency element string.

bin/parse_podfile_lock
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env ruby
+
+require 'json'
+require 'yaml'
+
+dependency_metadata_regex = /(?<pod>.*?)\s\((=|(~>))?\s?(?<version>(\d*\.?)+)\)/
+
+podfile_lock = nil
```
saikat056, line 8: We can write this as follows to avoid assigning `nil` to podfile_lock

```
podfile_lock = begin
  YAML.safe_load(File.read(ARGV[0]))
rescue SyntaxError, Psych::SyntaxError
  warn 'Unable to parse Podfile.lock file'
  exit(1)
end
```

Following is a quick test.
```
irb(main):001:1* a = begin
irb(main):002:1*   4
irb(main):003:1* rescue StandardError => e
irb(main):004:0> end
=> 4
irb(main):005:0> a
=> 4
irb(main):006:1* a = begin
irb(main):007:1*   raise StandardError.new
irb(main):008:1* rescue StandardError => e
irb(main):009:0> end
=> nil
irb(main):010:0> a
=> nil
```

bin/parse_podfile_lock
```diff
@@ -0,0 +1,43 @@
...
+  exit(1)
+end
+
+if podfile_lock['PODS'].nil? || podfile_lock['DEPENDENCIES'].nil?
+  warn '"PODS" or "DEPENDENCIES" key not found in Podfile.lock!'
+  exit(1)
+end
+
+dependencies = []
+# Parse individual dependency entry
```
saikat056, line 22: My understanding is that the bodies of the both loops are same. Can we write it in the following way? Correct me if I am wrong.

```
(podfile_lock['PODS'] + podfile_lock['DEPENDENCIES']).each do |dependency_entry|
  dependency_properties = dependency_metadata_regex.match(dependency_entry)
  dependency_hash = {}
  dependency_hash[:pod] = dependency_properties[:pod]
  dependency_hash[:version] = dependency_properties[:version]
  dependencies.append(dependency_hash)
end
```

spec/fixtures/report_podfile_lock/bad_podfile_nil_value/Podfile.lock
```diff
@@ -0,0 +1 @@
+null
```
saikat056, line 1: **Nit**: Please add a new line in order to make GitHub happy.

spec/fixtures/report_podfile_lock/empty_podfile_values/Podfile.lock
```diff
@@ -0,0 +1,7 @@
+PODS: 
+
+DEPENDENCIES: 
+
+PODFILE CHECKSUM: 
+
+COCOAPODS:
```
saikat056, line 7: **Nit**: Please add a new line in order to make GitHub happy.

spec/lib/salus/scanners/report_podfile_lock_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+
+  it 'should report an error when an unparseable file is found' do
+    repo = Salus::Repo.new('spec/fixtures/report_podfile_lock/empty_podfile_values')
+    scanner = Salus::Scanners::ReportPodfileLock.new(repository: repo, config: {})
+
+    scanner.run
+
+    errs = scanner.report.to_h.fetch(:errors)
+    expect(errs.size).to eq(1)
+    expect(errs[0][:message]).to eq('"PODS" or "DEPENDENCIES" key not found in Podfile.lock!
```
saikat056, line 89: 👏 👏 Thank you for writing this test for bin/parse_podfile_lock.

saikat056:
First of all, thank you for adding good test coverage to this pr. 🙏 

The change looks good to me. I added some minor suggestions. I want to see whether @ghbren or @joshua-ostrom has any other concerns. Otherwise, +1 from my side.

bin/parse_podfile_lock
ghbren: There is a `to_yaml` method here
https://www.rubydoc.info/github/CocoaPods/Core/Pod/Lockfile#to_yaml-instance_method

Gerald, could you look into whether you should use `to_yaml` or `YAML.safe_load`?

ghbren:
> First of all, thank you for adding good test coverage to this pr. 🙏
> 
> The change looks good to me. I added some minor suggestions. I want to see whether @ghbren or @joshua-ostrom has any other concerns. Otherwise, +1 from my side.

Thanks for the review.  I haven't looked at all the code, but I think the regex definitely needs to be replaced with parsing yaml before we merge.

bin/parse_podfile_lock
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env ruby
+
+require 'json'
+require 'yaml'
+
+podfile_lock = begin
+  YAML.safe_load(File.read(ARGV[0]))
+               rescue SyntaxError, Psych::SyntaxError
```
ghbren, line 8: `YAML.safe_load` throws more types of errors than the two above, like `Psych::DisallowedClass`, `Psych::BadAlias`, etc... 
It would be better to just `rescue StandardErrors`

lib/cyclonedx/report_podfile_lock_cyclonedx.rb
```diff
@@ -0,0 +1,10 @@
+module Cyclonedx
+  class ReportPodfileLock < Base
+    TYPE = "cocoa".freeze
```
ghbren, line 3: It appears that this `TYPE` is not used?

lib/cyclonedx/report_podfile_lock_cyclonedx.rb
```diff
@@ -0,0 +1,10 @@
...
+    UNLOCKED_DEPENDENCY_FILE = "Podfile".freeze
```
ghbren, line 4: ` UNLOCKED_DEPENDENCY_FILE` is also not used?

lib/cyclonedx/report_podfile_lock_cyclonedx.rb
geraldnash-cb: [It's used by the base class.](https://github.com/coinbase/salus/blob/8d4d7c8f80a0087b3e1fd1302be356d93369d842/lib/cyclonedx/base.rb#L42)

spec/lib/cyclonedx/report_podfile_lock_cyclonedx_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+      repo = Salus::Repo.new('spec/fixtures/report_podfile_lock/normal')
+
+      scanner = Salus::Scanners::ReportPodfileLock.new(repository: repo, config: {})
+      scanner.run
+
+      cyclonedx_report = Cyclonedx::Report.new([[scanner.report, false]],
+                                               { "spec_version" => "1.3" })
+      cyclonedx_report_hash = cyclonedx_report.to_cyclonedx
+
+      expect { Cyclonedx::Report.validate_cyclonedx(cyclonedx_report_hash) }.not_to raise_error
```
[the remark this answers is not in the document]
ghbren: Could you test what happens if with a lock file that cannot be parsed?
If you call `to_cyclonedx` on the result, will `validate_cyclonedx` still pass?

spec/lib/cyclonedx/report_podfile_lock_cyclonedx_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+        [
+          {
+            "bom-ref": "pkg:cocoa/boost-for-react-native@1.63.0",
+            type: "library",
+            group: "",
+            name: "boost-for-react-native",
+            version: '1.63.0',
+            purl: "pkg:cocoa/boost-for-react-native@1.63.0",
+            properties: [
+              { key: "source", value: "" },
```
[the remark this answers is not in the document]
joshua-ostrom: It would be nice to support the explicit source when given:

pod "Brightcove-Player-FreeWheel", :git => 'https://github.com/brightcove/BrightcoveSpecs.git'

Having that map to source would be a nice feature

spec/lib/cyclonedx/report_podfile_lock_cyclonedx_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+              { key: "dependency_file", value: "Podfile.lock" }
+            ]
+          },
+          {
+            "bom-ref": "pkg:cocoa/Flipper@%3D+0.87.0",
+            type: "library",
+            group: "",
+            name: "Flipper",
+            version: "= 0.87.0",
```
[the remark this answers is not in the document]
joshua-ostrom: Yes let's have version limited to the version, qualifiers should be removed

bin/parse_podfile_lock
```diff
@@ -0,0 +1,33 @@
...
+  exit(1)
+end
+
+dependencies = []
+# Parse individual dependency entry
+(podfile_lock['PODS'] + podfile_lock['DEPENDENCIES']).each do |dependency_entry|
+  dependency_properties = dependency_entry.split(' (')
+  dependency_hash = {}
+  dependency_hash[:pod] = dependency_properties[0]
+  dependency_hash[:version] = dependency_properties[1].delete_suffix(')')
```
joshua-ostrom, line 24: probably want to remove non numeric/alpha leading chars too.

bin/parse_podfile_lock
```diff
@@ -0,0 +1,37 @@
...
+# Checks the PODS and DEPENDENCIES section
+podfile_dependencies = (lockfile.send :pod_versions)
+
+normalized_dependencies = []
+podfile_dependencies.each do |dependency_entry|
+  dependency_hash = {}
+  dependency_hash[:pod] = dependency_entry[0]
+  dependency_hash[:version] = dependency_entry[1]
+  normalized_dependencies.append(dependency_hash)
+end
```
ghbren, line 30: Your code is fine here, but the can be updated to something like this to avoid accessing the array indexes.

```ruby
podfile_dependencies.each do |pod, version|
  dependency_hash = {}
  dependency_hash[:pod] = pod
  dependency_hash[:version] = version
  normalized_dependencies.append(dependency_hash)
end
```

ghbren (approved):

PR #538, opened 2022-02-02
Add npm package version scanning

Kamsiy:


lib/salus/scanners.rb
```diff
@@ -14,3 +14,10 @@ module Salus::Scanners; end
 
   require "salus/scanners/language_version/#{filename}"
 end
+
+Dir.entries(File.expand_path('scanners/package_version', __dir__)).sort.each do |filename|
```
joshua-ostrom, line 18: Now that we have 3 of these we should DRY it up.
Pseudo:

```
paths = ['scanners', 'scanners/language_version', 'scanners/language_version']
paths.each do |path|
  Dir.entries(File.expand_path(path, __dir__)).sort.each do |filename|
    next if ['.', '..'].include?(filename) # don't include FS pointers
    next unless /\.rb\z/.match?(filename)  # only include ruby files
    require "salus/#{path}/#{filename}"
  end
end
```

lib/sarif/sarif_report.rb
```diff
@@ -14,6 +14,11 @@
   require_relative "language_version/#{filename}"
 end
 
+Dir.entries(File.expand_path('./package_version', __dir__)).sort.each do |filename|
```
joshua-ostrom, line 17: DRY this up as well please

lib/salus/scanners/package_version/base.rb
```diff
@@ -0,0 +1,36 @@
...
+    def initialize(repository:, config:)
+      super
+      @passed = true
+    end
+
+    def run
+      # gather all package versions specified in scanner_config
+      package_versions = @config['package_versions']
+
+      # for each package in there, check if any violoations exist
```
joshua-ostrom, line 16: violations

lib/salus/scanners/package_version/npm_package_version_scanner.rb
```diff
@@ -0,0 +1,53 @@
...
+      super
+      @dependencies = should_run? ? generate_dependency_hash : {}
+    end
+
+    def should_run?
+      @repository.package_lock_json_present?
+    end
+
+    def check_for_violations(package_name, min_version, max_version)
+      if @dependencies.key?(package_name)
```
joshua-ostrom, line 15: This works... you could use npm list <PACKAGE NAME> too but I'm okay w/this 👍

joshua-ostrom (dismissed):

ghbren (approved):

PR #541, opened 2022-02-04
Add bower dependency reporter

maitrayshah:
Adding support for reporting on bower.json and .bowerrc file.


lib/salus/scanners/report_node_modules.rb
```diff
@@ -103,6 +110,36 @@ def record_dependencies_from_package_lock_json
...
+      # Record dependencies, version from bower.json
+      packages['dependencies']&.each do |name, version|
+        record_node_module(
+          name: name,
+          version: version,
+          source: sources,
+          dependency_file: 'bower.json'
+        )
+      end
+    rescue JSON::ParserError => e
```
ghbren, line 137: The `rescue JSON::ParseError` should only be applied to the two `JSON.parse(...)`, like
```ruby
begin
  file.JSON.parse(@repository.bower_json)
rescue JSON::ParseError ...
```

lib/salus/scanners/report_node_modules.rb
```diff
@@ -103,6 +110,36 @@ def record_dependencies_from_package_lock_json
...
+      # Record if custom registry is defined in .bowerrc
+      # Supports single and multiple registry -
+      # {"registry": "https://registry.bower.io"}
+      # {"registry": {"search": ["http://localhost:9000","https://registry.bower.io"] }}
+      sources = '<bower manager default>'
+      if @repository.bowerrc_present?
+        file = JSON.parse(@repository.bowerrc)
+        if file.key?("registry")
+          is_nested = file["registry"].is_a? String
+          sources = is_nested ? file["registry"] : file["registry"]["search"].join(',')
```
ghbren, line 124: 1. If `file["registry"]` is not a String, then is it always a hash?   If you are not sure, then maybe check if `is_a? Hash`.
2. Even if `registry` is a Hash, the `search` key could be missing, which would then lead to a `join` not defined on `nil` exception.

lib/salus/scanners/report_node_modules.rb
```diff
@@ -103,6 +110,36 @@ def record_dependencies_from_package_lock_json
       end
     end
 
+    def record_dependencies_from_bower_json
+      packages = JSON.parse(@repository.bower_json)
+      # Record if custom registry is defined in .bowerrc
+      # Supports single and multiple registry -
+      # {"registry": "https://registry.bower.io"}
+      # {"registry": {"search": ["http://localhost:9000","https://registry.bower.io"] }}
+      sources = '<bower manager default>'
```
ghbren, line 119: Is the actual `bower manager default` url defined in the bowser file?  It would be better to include an actual url instead of `bower manager default`.

spec/lib/cyclonedx/report_node_modules_cyclonedx_spec.rb
```diff
@@ -281,5 +281,89 @@
...
       )
     end
+
+    it 'should report all the deps in the bower.json if present' do
+      repo = Salus::Repo.new('spec/fixtures/report_node_modules/bower_json')
+      scanner = Salus::Scanners::ReportNodeModules.new(repository: repo, config: {})
+      scanner.run
+
+      node_cyclonedx = Cyclonedx::ReportNodeModules.new(scanner.report)
+      expect(node_cyclonedx.build_components_object).to match_array(
```
ghbren, line 291: Please check if the cyclonedx result can be validated against the schema.  Just need to make sure `validate_cyclonedx` doesn't throw an exception.
See https://github.com/coinbase/salus/blob/f0cb8ebc8743f020da5719d6da1c9779e4d11dc2/lib/cyclonedx/report.rb#L61

ghbren: Same for the other test case.

lib/salus/scanners/report_node_modules.rb
maitray16: Good point, let me add extra checks

ghbren (approved):

PR #545, opened 2022-02-08
Adding config option in ReportNodeModules to include/exclude dev dependencies

maitrayshah:
- Added configuration option to include/exclude dev dependencies for ReportNodeModules
```
- scanner_configs:
  ReportNodeModules:
   include_dev_deps: true
```

- Added cyclonedx 'scope' attribute for ReportNodeModules, default scope is 'required'. (https://cyclonedx.org/docs/1.2/json/#components_items_scope)
- Updated tests to handle new field introduced.
- Added test cases for handling new config and validating cyclonedx.


ghbren:
Please call the validate cyclonedex method in the specs.

lib/cyclonedx/report_node_modules_cyclonedx.rb
```diff
@@ -6,5 +6,12 @@ class ReportNodeModules < Base
     def initialize(scan_report)
       super(scan_report)
     end
+
+    def build_component(dependency)
```
saikat056, line 10: Can you please add the comment above the method definition instead of writing in the body? I think it would be easier for other developers to read the code.

```
# Add scope if available:
# https://cyclonedx.org/docs/1.3/json/#components_items_scope
def build_component(dependency)
   ....
end
```

lib/cyclonedx/report_node_modules_cyclonedx.rb
```diff
@@ -6,5 +6,13 @@ class ReportNodeModules < Base
     def initialize(scan_report)
       super(scan_report)
     end
+
+    # Add scope attribute, if available:
+    # https://cyclonedx.org/docs/1.3/json/#components_items_scope
+    def build_component(dependency)
+      super.merge({
+                    "scope": dependency[:scope] || "required"
```
joshua-ostrom, line 14: Minor nit, "required" is here and down in the reporting modules - can we place it in a const maybe over in the CycloneDX class and then reference that here and in the reporting modules?
SCOPE_REQUIRED = "required".freeze
SCOPE_OPTIONAL =  "optional".freeze

lib/salus/scanners/report_node_modules.rb
```diff
@@ -200,13 +206,14 @@ def record_dependencies_from_package_json(record_modules: true)
       end
     end
 
-    def record_node_module(dependency_file:, name:, version:, source:)
+    def record_node_module(dependency_file:, name:, version:, source:, scope: 'required')
```
joshua-ostrom, line 209: See prior note on const

lib/salus/scanners/report_node_modules.rb
```diff
@@ -101,12 +102,17 @@ def record_dependencies_from_package_lock_json
...
-        )
+        is_dev = data['dev'] || false
+
+        if !is_dev || is_dev && include_dev_dependencies
+          record_node_module(
+            name: dependency,
+            version: data['version'],
+            source: "#{data['resolved']}#{"##{data['integrity']}" if data['integrity']}",
+            dependency_file: 'package-lock.json',
+            scope: is_dev ? 'optional' : 'required'
```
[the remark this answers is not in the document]
joshua-ostrom: Given they are used elsewhere I'd vote of a const we can use for these 👍

joshua-ostrom (approved):

PR #550, opened 2022-02-15
Update CycloneDX Format

geraldnash-cb:
These changes update the shape of CycloneDX component JSON objects such that they can be consumed by a DependencyTrack instance. 

While the objects were CycloneDX-compatible before these changes (as evident by our CycloneDX schema validation tests), it was discovered that DependencyTrack would not accept components that contained keys such as `bom-ref` and `properties`. These changes make our CycloneDX compatible with DependencyTrack.


ghbren:
LGTM

joshua-ostrom:
A property on the cyclonedx report format config would be ideal for these.  Will approve but if we find a need to bring any back, defining their inclusion via a confg var would be nice.

joshua-ostrom (approved):

geraldnash-cb:
> A property on the cyclonedx report format config would be ideal for these. Will approve but if we find a need to bring any back, defining their inclusion via a confg var would be nice.

Noted!

PR #552, opened 2022-02-16
Recursive ReportNodeModules Scanner

maitrayshah:
- Add nested dependencies support for ReportNodeModules Scanner.
- Respect `include_dev_deps` flag for both direct and nested dependency.
- Updating test cases to account for nested dependencies.


joshua-ostrom (approved):
Thanks for fixing this!

PR #557, opened 2022-02-22
Ensure we expand the path

joshua-ostrom:
Small fix to ensure avoid issues with nesting relative paths


spec/lib/salus/processor_spec.rb
```diff
@@ -7,6 +7,10 @@ def remove_key(json_string, key = 'running_time')
     json['scans'].each do |scanner, _|
       json['scans'][scanner].delete(key)
     end
+    # Avoid comparing relative and absolute file:///
+    json['config']['report_uris'].each_with_index do |endpoint, index|
```
ghbren, line 11: Is it possible that the json has no `config` key, or `json['config']` has no 'report_uris' keys?

joshua-ostrom: I can add a dig, to future proof it in case any adds some specs fixtures that lack a report endpoint.  That's unlikely as the method here is used to compare the output vs the expected output but future proofing it sounds good.

ghbren: Got it, no need to add the dig then.

ghbren (dismissed):

ghbren (dismissed):

ghbren (approved):

PR #558, opened 2022-02-22
OSV Scanner

maitrayshah:
- Added: Support for handling connection to [OSV](https://osv.dev/) and exceptions config.
- Added: GoOSV scanner to parse dependencies and comparing with OSV data.
- Added: Test cases for GoOSV scanner.


lib/salus/scanners/osv/base.rb
```diff
@@ -0,0 +1,133 @@
...
+              "RubyGems/all.zip"
+            elsif @repository.cargo_present? || @repository.cargo_lock_present?
+              "https://osv-vulnerabilities.storage.googleapis.com/"\
+              "crates.io/all.zip"
+            elsif @repository.package_lock_json_present?
+              "https://osv-vulnerabilities.storage.googleapis.com/"\
+              "npm/all.zip"
+            end
+      URI(url)
+    end
```
saikat056, line 53:
```
def method_a
    return x if b == 1
    return y if b == 2
    ...
end
```
The above pattern is preferable to if..else in ruby. Search "Prefer guard clause" in https://confluence.coinbase-corp.com/pages/viewpage.action?pageId=823230474

We can write the code as follows:

```
def osv_url
   return osv_url_for("Go") if go_repo?
   return ...  if  ...
   return ...  if  ...
end

private

def go_repo?
  @repository.go_sum_present? || @repository.go_mod_present?
end

def osv_url_for(package)
   url = "https://osv-vulnerabilities.storage.googleapis.com/"
            + package + "all.zip"
   URI(url)
end
```

WDYT? 🤔

maitrayshah: Yeah that makes sense, let me update that.

lib/salus/scanners/osv/base.rb
```diff
@@ -0,0 +1,131 @@
...
+
+    def should_run?
+      raise NoMethodError
+    end
+
+    private
+
+    def name
+      self.class.name.sub('Salus::Scanners::OSV::', '')
```
ghbren, line 24: What is this method supposed to do?
My guess is that if `self.class` is `A::B::C`, then `name` should return `"C"`?
If so, you might want to call `self.class.name.split('::')[-1]` instead, so that you don't need hardcode the class name (which could change).

lib/salus/scanners/osv/base.rb
```diff
@@ -0,0 +1,131 @@
...
+require 'net/http'
+
+module Salus::Scanners::OSV
+  class Base < Salus::Scanners::Base
+    class ApiTooManyRequestsError < StandardError; end
+
+    MAX_RETRIES = 2
+
+    def run
+      raise NoMethodError
```
ghbren, line 14: Might be better to add a message here like
```ruby
      raise NoMethodError, 'implement in subclass'
```

ghbren: same for `should_run?`

lib/salus/scanners/osv/base.rb
```diff
@@ -0,0 +1,131 @@
...
+
+    def name
+      self.class.name.sub('Salus::Scanners::OSV::', '')
+    end
+
+    def osv_vulnerabilities
+      @osv_vulnerabilities ||= fetch_vulnerabilities
+    end
+
+    def osv_url
```
ghbren, line 31: This method will only return the url for a single language.  
But a repo may have multiple languages, like repo has go.sum, requirements.txt,  and Gemfile.lock.

This method should instead return urls for all applicable languages, right?

lib/salus/scanners/osv/base.rb
```diff
@@ -0,0 +1,131 @@
...
+          str += "#{k}: #{v}\n"
+        end
+        str += "\n"
+      end
+      str
+    end
+
+    def send_request
+      vulns = []
+      response = Net::HTTP.get_response(osv_url)
```
ghbren, line 113: What if there is no supported language?  Will `osv_url` return `nil` or empty string or something?

ghbren:
Could you paste a sample output txt from `docker run ...`?

lib/salus/scanners/osv/base.rb
```diff
@@ -0,0 +1,131 @@
...
+        sleep rand(0..max_sleep_seconds)
+        retry
+      else
+        msg = "Too many requests to OSV url after " \
+        "#{retries} retries"
+        report_error("Connection to OSV failed: #{msg}")
+      end
+    end
+
+    def flatten_by_affected(doc)
```
ghbren, line 90: I'm a little confused about this method.
Could you add a little documentation on this method with sample input?

lib/salus/scanners/osv/base.rb
```diff
@@ -0,0 +1,131 @@
...
+
+      # Group vulnerabilities by aliases / id
+      grouped = all_vulnerabilities_found.group_by { |d| d.fetch("aliases", [d.fetch("id")]) }
+      grouped.each do |_key, value|
+        results.append(value[0])
+      end
+      results
+    rescue StandardError => e
+      report_error("Connection to OSV failed: #{e}")
+    rescue ApiTooManyRequestsError
```
ghbren, line 77: This `rescue ApiTooManyRequestsError` will never execute, right?  `ApiTooManyRequestsError` is a subclass of `StandardError`.  So whenever there is an `ApiTooManyRequestsError`, then only the `rescue StandardError` above will execute.

ghbren: So I think you need to move `rescue ApiTooManyRequestsError` above 'rescue StandardError`.

lib/salus/scanners/osv/base.rb
```diff
@@ -0,0 +1,131 @@
...
+      if retries < MAX_RETRIES
+        retries += 1
+        max_sleep_seconds = Float(2**retries)
+        sleep rand(0..max_sleep_seconds)
```
ghbren, line 81: Any reason for sleeping for a random time instead of just
`sleep max_sleep_seconds`?

lib/salus/scanners/osv/base.rb
maitrayshah: Yup, I don't think I use it so can remove it.

lib/salus/scanners/osv/base.rb
```diff
@@ -0,0 +1,131 @@
...
+      # Group vulnerabilities by aliases / id
+      grouped = all_vulnerabilities_found.group_by { |d| d.fetch("aliases", [d.fetch("id")]) }
+      grouped.each do |_key, value|
+        results.append(value[0])
+      end
+      results
+    rescue StandardError => e
+      report_error("Connection to OSV failed: #{e}")
+    rescue ApiTooManyRequestsError
+      if retries < MAX_RETRIES
```
ghbren, line 78: I don't see where `retries` is initialized. Did I miss something?

lib/salus/scanners/osv/base.rb
maitrayshah: So the results returned from OSV are nested by packages, so I flatten them to make it easier to compare dependencies. I add paste an example to make it easier.

lib/salus/scanners/osv/base.rb
maitrayshah: Good catch, I thought I had it in the correct order.

ghbren:
Any rate limits on pulling the vulnerabilities from osv-vulnerabilities?

lib/salus/scanners/osv/base.rb
maitrayshah: I was using implementation already in one of the classes but we can use a fixed time.

lib/salus/scanners/osv/go_osv.rb
```diff
@@ -0,0 +1,119 @@
...
+          )
+        end
+      end
+
+      # Pick specific version of dependencies
+      # If multiple versions of dependencies are found then pick the max version to mimic MVS
+      # https://go.dev/ref/mod#minimal-version-selection
+      all_dependencies.each do |deps|
+        lib = deps[:name]
+        version = deps[:version].to_s.gsub('v', '').gsub('+incompatible', '')
```
ghbren, line 44: Could there be an edge case where `deps[:version]` looks like `v1.2.3vxyz`, which would break the line above?

Would it work if you just get rid of the first character if it's `v`?

maitrayshah: Yeah I think switching to just replacing first char if it's 'v' that might be better.

lib/salus/scanners/osv/base.rb
maitrayshah: I think since the subclass implements should_run? check, so it would be covered by it. For example - GoOSV checks for go.sum file presence if not the scanner doesn't run, so the nil should not be an issue. 
Let me know what you think ?

lib/salus/scanners/osv/base.rb
```diff
@@ -0,0 +1,138 @@
...
+      urls.append(osv_url_for("Go")) if @repository.go_sum_present? || @repository.go_mod_present?
+      urls.append(osv_url_for("PyPI")) if @repository.requirements_txt_present?
+      urls.append(osv_url_for("Maven")) if @repository.pom_xml_present?
+
+      urls
+    end
+
+    def osv_url_for(package)
+      url = "https://osv-vulnerabilities.storage.googleapis.com/"\
+      + package + "/all.zip"
```
joshua-ostrom, line 39: Can you document at this time the list of packages supported and their size?

go/all.zip 82KB
etc etc

lib/salus/scanners/osv/go_osv.rb
```diff
@@ -0,0 +1,124 @@
...
+
+    EMPTY_STRING = "Not Found".freeze
+    DEFAULT_SOURCE = "https://osv.dev/list".freeze
+    DEFAULT_SEVERITY = "LOW".freeze
+
+    def should_run?
+      @repository.go_sum_present?
+    end
+
+    def find_dependencies
```
joshua-ostrom, line 16: Let's DRY up our code here.  We have existing language dependency reporters - each of which has logic for parsing the dependencies present.  I'd like to see us have those reporters and these OSV scanners use a common parsing method.

For example:
https://github.com/coinbase/salus/blob/master/lib/salus/scanners/report_go_dep.rb

Let's refactor as needed so we can have one common parser for each package manager.

lib/salus/scanners/osv/go_osv.rb
```diff
@@ -0,0 +1,124 @@
...
+        return report_error("GoOSV: #{msg}")
+      end
+
+      # Match dependencies found with advisories from Github
+      if osv_vulnerabilities.nil?
+        msg = "No vulnerabilities found to compare."
+        bugsnag_notify("GoOSV: #{msg}")
+        return report_error("GoOSV: #{msg}")
+      else
+        dependencies.each do |lib, version|
```
joshua-ostrom, line 71: Minor nit: Can you move this block to a method for readability
```
def find_vulns_present(deps, osv_vulns)
   ...
   results
end
```
Or similar

lib/salus/scanners/osv/go_osv.rb
maitrayshah: One of the options I am thinking is following the similar method to what we use, adding a parse_go_sum file bin/ folder.
This way both report_go_dep and go_osv can call that to get dependencies.

maitrayshah:
> Could you paste a sample output txt from `docker run ...`?

```
[INFORMATIONAL: 2022-02-25 01:43:38 +0000]: Salus is running.
GoOSV is scanning...
GoOSV has finished
==== Salus Scan v2.17.0

==== GoOSV: FAILED in 1.21s

 ~~ Scanner Logs:

  Package: github.com/satori/go.uuid
  Vulnerable Version: 0
  Version Detected: 1.2.0
  Patched Version: 1.2.1.pre.0.20181016170032.pre.d91630c85102
  ID: GO-2020-0018
  Summary: UUIDs generated using [`NewV1`] and [`NewV4`] may not read the expected
  number of random bytes. These UUIDs may contain a significantly smaller
  amount of entropy than expected, possibly leading to collisions.
  References: https://github.com/satori/go.uuid/pull/75, https://github.com/satori/go.uuid/commit/d9
  1630c8510268e75203009fe7daf2b8e1d60c45, https://github.com/satori/go.uuid/issues/73
  Source: https://go.googlesource.com/vulndb/+/refs/heads/master/reports/GO-2020-0018.yaml
  Severity: LOW
```

lib/salus/scanners/osv/base.rb
```diff
@@ -0,0 +1,152 @@
...
+      self.class.name.sub('Salus::Scanners::OSV::', '')
+    end
+
+    private
+
+    def osv_vulnerabilities
+      @osv_vulnerabilities ||= fetch_vulnerabilities
+    end
+
+    def osv_url
```
saikat056, line 27: def osv_url -> def osv_urls

Because, the method is returning an array of urls

maitrayshah: Good point

lib/salus/scanners/osv/base.rb
```diff
@@ -0,0 +1,152 @@
...
+        end
+      end
+
+      # Group vulnerabilities by aliases / id
+      # Picks 1 item from each grouped item
+      # { "ID-1": [
+      #      {"name": "sample-dep-1"},  {"name": "sample-dep-1"}
+      #    ]
+      # }
+      grouped = all_vulnerabilities_found.group_by { |d| d.fetch("aliases", [d.fetch("id")]) }
```
saikat056, line 75:
```
results = []
grouped = all_vulnerabilities_found.group_by { |d| d.fetch("aliases", [d.fetch("id")]) }
```
Please declare `results` close to it's first usage.

maitrayshah: Gonna update this wherever I declare variables at the top.

lib/salus/scanners/osv/go_osv.rb
```diff
@@ -0,0 +1,128 @@
...
+      end
+
+      if osv_vulnerabilities.nil?
+        msg = "No vulnerabilities found to compare."
+        bugsnag_notify("GoOSV: #{msg}")
+        return report_error("GoOSV: #{msg}")
+      else
+        results = match_vulnerable_dependencies(dependencies)
+      end
+      # Report scanner status
```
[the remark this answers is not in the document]
maitrayshah: Good point, definitely will keep in mind.

lib/salus/scanners/osv/go_osv.rb
```diff
@@ -0,0 +1,128 @@
...
+        report_stderr(err_msg)
+        report_error(err_msg)
+        return
+      end
+
+      dependencies = {}
+      # Pick specific version of dependencies
+      # If multiple versions of dependencies are found then pick the max version to mimic MVS
+      # https://go.dev/ref/mod#minimal-version-selection
+      all_dependencies.each do |deps|
```
saikat056, line 64: `all_dependencies.each do |deps|` -> `all_dependencies.each do |dependency|`

`dep` is a single dependency. Am I right? 🤔

maitrayshah: Yup, its single dependency. Good catch

spec/lib/sarif/osv/go_osv_sarif_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+    let(:repo) { Salus::Repo.new('spec/fixtures/osv/go_osv/failure_vulnerability_present') }
+
+    it 'should generate an empty sarif report' do
+      scanner = Salus::Scanners::OSV::GoOSV.new(repository: repo,
+        config: scanner_config)
+      stub_req_with_valid_response
+      scanner.run
+      report = Salus::Report.new(project_name: "Neon Genesis")
+      report.add_scan_report(scanner.report, required: false)
+      report_object = JSON.parse(report.to_sarif)['runs'][0]
```
saikat056, line 83: Please add a new line after line 83.

lib/salus/scanners/osv/go_osv.rb
```diff
@@ -0,0 +1,128 @@
...
+        else
+          dependencies[lib] = version
+        end
+      end
+      dependencies
+    end
+
+    # Compare vulnerabilities found with dependencies found
+    # and return vulnerable dependencies
+    def match_vulnerable_dependencies(dependencies)
```
saikat056, line 78: The method `match_vulnerable_dependencies` has become very long. Is it possible to extract some parts for it to separate smaller methods so that we can make it smaller. 🤔

maitrayshah: Let me look into it, i was thinking about it but wasn't sure if there was a clean way.

maitrayshah: Moving version checking to it own function so that should make it much better.

bin/parse_go_sum
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env ruby
+
+require 'json'
+
+parsed_file = nil
+begin
+  parsed_file = File.read(ARGV[0])
+rescue StandardError
+  warn 'Unable to read and parse go.sum file'
```
saikat056, line 9: Is it possible to write a unit-test for this warning?

[Here](https://github.com/coinbase/salus/blob/aad632b77a5fbd20af302f3f08dd53d8dba6e2ea/spec/lib/salus/scanners/report_podfile_lock_spec.rb#L77) is an example.

maitrayshah: Yup, will add

maitrayshah: Actually, might not be needed to rescue here since 'empty' and 'missing' go.sum file is being handled by the calling class, so should be safe to remove from here. Will update accordingly.

lib/salus/scanners/osv/base.rb
```diff
@@ -0,0 +1,148 @@
...
+      urls = []
+      urls.append(osv_url_for("Go")) if @repository.go_sum_present? || @repository.go_mod_present?
+      urls.append(osv_url_for("PyPI")) if @repository.requirements_txt_present?
+      urls.append(osv_url_for("Maven")) if @repository.pom_xml_present? ||
+        @repository.build_gradle_present?
+
+      urls
+    end
+
+    def osv_url_for(package)
```
joshua-ostrom, line 45: At those sizes we should probably spin up a caching layer (given the volume of daily scans).

Let's create create a plugin hook here so we can easily override the loading of vulnerabilities as well as the matching of vulnerabilities.   This will allow salus to ship with a direct call (like you have here) and provide an easy override to cache locally.    Medium term we will plan on a normalized data layer for Salus - but we can tackle that w/out blocking this PR.

bin/parse_go_sum
```diff
@@ -0,0 +1,27 @@
...
+      {
+        "namespace" => names[0..-2].join("/").to_s,
+        "name" => names[-1].to_s,
+        "version" => parts[1].to_s.gsub('/go.mod', ''),
+        "checksum" => parts[2].to_s.gsub('h1:', '')
+      }
+    )
+  end
+end
+
```
ghbren, line 26: If I'm not mistaken, shell_return looks for specific status exit codes, so I think you should add the exit codes here.

bin/parse_go_sum
```diff
@@ -0,0 +1,27 @@
...
+  # Validate 3 parts - name, version and checksum
+  parts = line.split
+  if parts.length == 3
+    names = parts[0].split("/")
+    dependencies.append(
+      {
+        "namespace" => names[0..-2].join("/").to_s,
+        "name" => names[-1].to_s,
+        "version" => parts[1].to_s.gsub('/go.mod', ''),
+        "checksum" => parts[2].to_s.gsub('h1:', '')
```
ghbren, line 21: what is `h1`? Can it ever have non-h1 values?

ghbren: the checksum line will have issues if the checksum itself contains `h1`.

bin/parse_go_sum
ghbren: Is there an existing library that parses go dependency files like this?

lib/sarif/osv/base_sarif.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    def initialize(scan_report, repo_path = nil)
+      super(scan_report, {}, repo_path)
+      @uri = OSV_URI
+      @logs = parse_scan_report!
+    end
+
+    def parse_scan_report!
+      logs = []
+      advisories = @scan_report.to_h.dig(:logs).split("\n\n")
```
ghbren, line 18: Are you parsing plain text here?  Is there a way to get the logs to show up as json or something?

bin/parse_go_sum
maitrayshah: Let me add that.

bin/parse_go_sum
maitrayshah: Not that I have seen, h1: is always there - 
Example - `cloud.google.com/go v0.22.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=`

maitrayshah: Also, won't the gsub just match `"h1:"`, so `h1` in the checksum shouldn't be affected

lib/salus/scanners/osv/base.rb
maitrayshah: Tracking internally.

lib/salus/scanners/osv/base.rb
```diff
@@ -0,0 +1,137 @@
...
+      # PyPI: all.zip (3.4 MB) -> 10.9 MB
+      # Maven: all.zip (1.6 MB) -> 5.1 MB
+      url = "https://osv-vulnerabilities.storage.googleapis.com/"\
+      + package + "/all.zip"
+      URI(url)
+    end
+
+    def fetch_vulnerabilities
+      # Flatten vulnerabilities by package name.
+      all_vulnerabilities = send_request
```
ghbren, line 58: but `send_request` could raise an exception. That exception should be checked here.

lib/salus/scanners/osv/base.rb
```diff
@@ -0,0 +1,137 @@
...
+      end
+      all_vulnerabilities_found.flatten!
+
+      # Handle removing exception ids from vulnerabilities found.
+      exception_ids = fetch_exception_ids
+      if exception_ids.any? && all_vulnerabilities_found.any?
+        all_vulnerabilities_found.delete_if do |v|
+          identifiers_found = v.fetch("aliases", []) + [v.fetch("id")]
+          intersection = identifiers_found & exception_ids
+          true if intersection.length.positive?
```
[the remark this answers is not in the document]
maitrayshah: Good catch.

bin/parse_go_sum
```diff
@@ -0,0 +1,31 @@
...
+      }
+    )
+  end
+end
+
+puts dependencies.to_json
```
joshua-ostrom, line 31: might be handy to have a high level structure here:
```
{
	"dependencies": {
		"parsed": [{
			"namespace": "",
			"name": "",
			"version": "",
			"checksum": ""
		}],
		"skipped": []
	}
}
``` That would afford some flexibility if we wanted to track or log any deps we couldn't parse

File: lib/salus/scanners/osv/base.rb [hunk 28]
```diff
@@ -0,0 +1,137 @@
...
+require 'net/http'
+
+module Salus::Scanners::OSV
+  class Base < Salus::Scanners::Base
+    DATABASE_STRING_MAPPING = {
+      "GHSA" => "Github Advisory Database",
+      "PYSEC" => "Python Packaging Advisory Database",
+      "GO" => "Go Vulnerability Database",
+      "RUSTSEC" => "RustSec Advisory Database",
+      "default" => "Open Source Vulnerabilitiy"
```
joshua-ostrom, line 14: Do the OSV ones lack a prefix?  It might be nice to capture unsupported ones too.  We could then setup monitors for "unsupported" entries that would alert us when OSV adds a new datasource and wee need to update the mapping here.  Thoughts?

lib/salus/scanners/osv/base.rb
```diff
@@ -0,0 +1,137 @@
...
+
+    private
+
+    def osv_vulnerabilities
+      @osv_vulnerabilities ||= fetch_vulnerabilities
+    end
+
+    def osv_urls
+      urls = []
+      urls.append(osv_url_for("Go")) if @repository.go_sum_present? || @repository.go_mod_present?
```
joshua-ostrom, line 37: if @repository.go_sum_present? || @repository.go_mod_present? differers from the Go osv scanner should_run? logic.

As setup this will download all of the OSV datasets that the repo has lock files for.  The repo config, however may not have a given scanner enabled (pypi osv scanner for example).  To avoid this, and DRY up the inclusion logic, what do you think about having the subclass invoke a method to request the data pull?

bin/parse_go_sum
maitrayshah: Yup this makes sense, let me add that.

lib/salus/scanners/osv/base.rb
maitrayshah: Using default: "Open Source Vulnerability as the "unsupported" catch all. Since OSV pulls from other sources so if that is not mapped I default to "Open Source Vulnerability".

lib/salus/scanners/osv/go_osv.rb
```diff
@@ -0,0 +1,150 @@
...
+        return
+      end
+
+      @osv_vulnerabilities ||= fetch_vulnerabilities(GO_OSV_ADVISORY_URL)
+      results = []
+      if @osv_vulnerabilities.nil?
+        msg = "No vulnerabilities found to compare."
+        bugsnag_notify("GoOSV: #{msg}")
+        return report_error("GoOSV: #{msg}")
+      else
```
joshua-ostrom, line 32: you can drop the else

joshua-ostrom (approved):

joshua-ostrom (approved):

PR #562, opened 2022-03-03
MavenOSV Scanner

maitrayshah:
- Adding support for Maven CVE Scanning.
- Updating `report_pom_xml` to add versions found in`<properties>` tag.
- Adding SARIF support.


bin/parse_pom_xml
```diff
@@ -22,17 +22,28 @@ if parsed_file.errors.size.positive?
   exit(1)
 end
 
-dependencies = []
+properties = {}
+# Parse contents within each <properties> tag
+nodes = parsed_file.css('properties')
```
ghbren, line 27: If `properties` does not exist, then will `nodes` be empty array or `nil`?
If `nil`, then the `node.each` will throw an exception.

bin/parse_pom_xml
```diff
@@ -22,17 +22,28 @@ if parsed_file.errors.size.positive?
...
+nodes.each do |node|
+  childrens = node.children
+  childrens.each do |children|
```
ghbren, line 30: This is 2 levels of nesting.  Could there be more than 2?

bin/parse_pom_xml
```diff
@@ -22,17 +22,28 @@ if parsed_file.errors.size.positive?
...
+  end
+end
 
+dependencies = []
 # Parse contents within each <dependency> tag
 parsed_file.css('dependency').each do |dependency|
   group_id = dependency.at('groupId')
   artifact_id = dependency.at('artifactId')
   version = dependency.at('version')
+  version = properties.dig(version.children.to_s) || version.children.to_s unless version.nil?
```
ghbren, line 41: could you provide a pom example of why this is needed?

lib/salus/scanners/osv/maven_osv.rb
```diff
@@ -0,0 +1,150 @@
+require 'salus/scanners/osv/base'
+
+module Salus::Scanners::OSV
+  class MavenOSV < Base
+    class SemVersion < Gem::Version; end
+
+    EMPTY_STRING = "Not Found".freeze
```
ghbren, line 7: could you set it to `""` to be consistent with the other scanners?

lib/salus/scanners/osv/maven_osv.rb
```diff
@@ -0,0 +1,150 @@
...
+    DEFAULT_SOURCE = "https://osv.dev/list".freeze
+    DEFAULT_SEVERITY = "MODERATE".freeze
```
ghbren, line 9: Does OSV have "MEDIUM" as a severity?  If so, could you set the default to "MEDIUM", just to be consistent with our other tools?

lib/salus/scanners/osv/maven_osv.rb
```diff
@@ -0,0 +1,150 @@
...
+    GITHUB_DATABASE_STRING = "Github Advisory Database".freeze
```
ghbren, line 10: Are there any other scanners that use `GITHUB_DATABASE_STRING`?  If so, maybe move it to a super class?

lib/salus/scanners/osv/maven_osv.rb
```diff
@@ -0,0 +1,150 @@
...
+        err_msg = "MavenOSV: Failed to parse any dependencies from the project."
+        report_stderr(err_msg)
+        report_error(err_msg)
+        return
+      end
+
+      @osv_vulnerabilities ||= fetch_vulnerabilities(MAVEN_OSV_ADVISORY_URL)
+      if @osv_vulnerabilities.nil?
+        err_msg = "No vulnerabilities found to compare."
+        bugsnag_notify("MavenOSV: #{err_msg}")
```
ghbren, line 30: It would be good to `report_stderr` too.

ghbren, line 30: `report_error` calls `bugsnag_notify`, so there is no need to call `bugsnag_notify` separately.
But you could update `err_msg` to `MavenOSV: No vulnerabilities found to compare."

lib/salus/scanners/osv/maven_osv.rb
```diff
@@ -0,0 +1,150 @@
...
+        return report_error("MavenOSV: #{err_msg}")
```
ghbren, line 31: It would be good to `report_stderr` too. Otherwise the error won't show in the txt output.

ghbren:
Have you tested this PR by building a local docker container, then `docker run`?
Also, update salus.yaml to write to sarif?

lib/salus/scanners/osv/maven_osv.rb
```diff
@@ -0,0 +1,150 @@
...
+      end
+
+      # Fetch vulnerable dependencies.
+      # Dedupe and select Github Advisory over other sources if available.
+      results = []
+      grouped = match_vulnerable_dependencies(dependencies).group_by { |d| d[:ID] }
+      grouped.each do |_key, values|
+        vuln = {}
+        values.each do |v|
+          vuln = v if v[:Database] == GITHUB_DATABASE_STRING
```
ghbren, line 41: Are other databases out of the scope for this PR?

maitrayshah: Not out of scope but sometimes there are multiple advisories coming from different databases. So we try to pick Github Advisory over other ones and also helps with deduplication.

spec/lib/sarif/osv/maven_osv_sarif_spec.rb
```diff
@@ -0,0 +1,137 @@
+require_relative '../../../spec_helper'
+require 'json'
+
+describe Sarif::MavenOSVSarif do
```
ghbren, line 4: Please call `validate_sarif` on the result sarifs. You can grep for it to see how it's used in other specs.

maitrayshah:
> 

Yup test it with following config - 
```
active_scanners:
  - MavenOSV

reports:
  - uri: file://salus-report.sarif
    format: sarif
```

maitrayshah:
> Have you tested this PR by building a local docker container, then `docker run`? Also, update salus.yaml to write to sarif?
Yup test it with following config -

active_scanners:
  - MavenOSV

reports:
  - uri: file://salus-report.sarif
    format: sarif

lib/salus/scanners/osv/maven_osv.rb
maitrayshah: The super class adds a `database` field depending on ID and the sub class cherry picks `GITHUB_DATABASE_STRING`. The reason I added in sub class is to allow sub class to pick advisories they want and the super class returns everything.

lib/salus/scanners/osv/maven_osv.rb
maitrayshah: Severity labels are - CRITICAL, HIGH, MODERATE, LOW

lib/salus/scanners/osv/maven_osv.rb
```diff
@@ -0,0 +1,150 @@
...
+    def match_vulnerable_dependencies(dependencies)
+      results = []
+      dependencies.each do |dependency|
+        lib = "#{dependency['group_id']}:#{dependency['artifact_id']}"
+
+        unless dependency['version'].nil?
+          version = dependency['version']
+          # If version is of the format '${deps.version}', log it.
+          if version.match(/\${(.*)\.version}/)
+            bugsnag_notify("MavenOSV: Found #{lib}:#{version} with incompatible format.")
```
ghbren, line 89: Does `${deps.version}` indicate the pom is mis-configured?

lib/salus/scanners/osv/maven_osv.rb
```diff
@@ -0,0 +1,150 @@
...
+            break
```
ghbren, line 90: Did you mean `break` or `next`? because `break` would skp the rest of the dependencies?

bin/parse_pom_xml
maitrayshah: If `<properties>` is absent then returns an empty array so should be good.

lib/salus/scanners/osv/maven_osv.rb
```diff
@@ -0,0 +1,150 @@
...
+
+      vulnerable_flag
+    end
+
+    # Compare vulnerabilities found with dependencies found
+    # and return vulnerable dependencies
+    def match_vulnerable_dependencies(dependencies)
+      results = []
+      dependencies.each do |dependency|
+        lib = "#{dependency['group_id']}:#{dependency['artifact_id']}"
```
ghbren, line 83: are `group_id` and `artifact_id` keys that always exist?

bin/parse_pom_xml
maitrayshah: In pom.xml, dependencies are sometimes defined as - 
```
<dependency>
    <groupId>ch.qos.logback</groupId>
    <artifactId>logback-classic</artifactId>
    <version>${logback.version}</version>
</dependency>
```
In this case we used to report version as `${logback.version}` which is incorrect and can cause issues. The code highlighted is a fix for it.
These variables are defined in `<properties>` tag
```
<properties>
<logback.version>1.2.10</logback.version>
</properties>
```
So I parse it and try to map it when building the dependencies array to get more accurate representation.

lib/salus/scanners/osv/maven_osv.rb
```diff
@@ -0,0 +1,150 @@
...
+            bugsnag_notify("MavenOSV: Found #{lib}:#{version} with incompatible format.")
+            break
+          end
+
+          package_matches = @osv_vulnerabilities.select do |v|
+            v.dig("package", "name") == lib
+          end
+          package_matches.each do |m|
+            m["ranges"].each do |version_ranges|
+              if version_ranges["type"] == "SEMVER" || version_ranges["type"] == "ECOSYSTEM"
```
ghbren, line 98: should we be skipping all `version_ranges` that do not match the above?

lib/salus/scanners/osv/maven_osv.rb
```diff
@@ -0,0 +1,150 @@
...
+                introduced = version_ranges["events"][0]["introduced"]
+                fixed = if version_ranges["events"].length == 2
+                          version_ranges["events"][1]["fixed"]
```
ghbren, line 101: do `introduced` and `fixed` always exist as keys?

maitrayshah: introduced always exist, `fixed` doesn't that's why the check for length == 2. So both `introduced` and `fixed` will exist.

lib/salus/scanners/osv/maven_osv.rb
```diff
@@ -0,0 +1,150 @@
...
+
+    # Match if dependency version found is in the range of
+    # vulnerable dependency found.
+    def version_matching(version, version_ranges)
+      vulnerable_flag = false
+      version_found = SemVersion.new(version)
+      # If version range length is 1, then no fix available.
+      if version_ranges.length == 1
+        introduced = SemVersion.new(
+          version_ranges[0]["introduced"]
```
ghbren, line 61: do `introduced` and `fixed` always exist as keys?

lib/salus/scanners/osv/maven_osv.rb
maitrayshah: For now I have encountered 3 types - SEMVER, ECOSYSTEM, GIT.
So we ignore the GIT ones since it lists which git commit introduced the vulnerability. We don't have a way to compare git commits from the dependency list and also it gets covered under SEMVER, ECOSYSTEM.

lib/salus/scanners/osv/maven_osv.rb
ghbren, line 61: if length is 1, could it the `introduced` key still not exist?

lib/salus/scanners/osv/maven_osv.rb
```diff
@@ -0,0 +1,150 @@
...
+        )
+        vulnerable_flag = true if version_found >= introduced
+      # If version range length is 2, then both introduced and fixed are available.
+      elsif version_ranges.length == 2
```
ghbren, line 65: if length is 2, then do `introduced` and 'fixed` keys always exist.

lib/salus/scanners/osv/maven_osv.rb
ghbren: Ok, then we can leave it as MODERATE for now.

lib/salus/scanners/osv/maven_osv.rb
maitrayshah: Not misconfigured but we aren't able to match a version for it. With the properties parsing introduced I don't think it should be triggered but still adding in just in case if we encounter it.

lib/salus/scanners/osv/maven_osv.rb
ghbren: But is it possible for length to be 2, but the keys are not introduced/keys?

bin/parse_pom_xml
maitrayshah: Nope, so should be good here.

lib/salus/scanners/osv/maven_osv.rb
maitrayshah: Could be (for example - if the input we get is bad) but I haven't seen an example of it since they are following the OSV Format. So according the format these fields would need to be there.

lib/salus/scanners/osv/maven_osv.rb
ghbren: Ok. Hopefully we don't suddenly get a project with tons of `${deps.version}` and then get tons of bugsnag pings.

But should you instead check for anything that start with `$`, they could put other variable names inside, right?

lib/salus/scanners/osv/maven_osv.rb
maitrayshah: Same response as above - 
Could be (for example - if the input we get is bad) but I haven't seen an example of it since they are following the OSV Format. So according the format these fields would need to be there.

lib/salus/scanners/osv/maven_osv.rb
ghbren: I wonder if there is a tool that gives the actual version of `${deps.version}`. Though may be outside the scope of this PR.

maitrayshah: There might be but I am not too worried about it since they won't show up under `<dependencies>` tag. So the main thing I saw which we needed to handle was `${deps.version}` which gets addressed while parsing `pom.xml ` and added a check in case if we encounter any issues.

lib/salus/scanners/osv/maven_osv.rb
maitrayshah: Yes

bin/parse_pom_xml
joshua-ostrom: could you add that sample xml into the code as a comment?

lib/salus/scanners/osv/maven_osv.rb
```diff
@@ -0,0 +1,152 @@
...
+              end
+            end
+          end
+        end
+      end
+      results
+    end
+
+    # Find dependencies from the project
+    def find_dependencies
```
joshua-ostrom, line 132: This block is a copy of https://github.com/coinbase/salus/blob/master/lib/salus/scanners/report_pom_xml.rb#L10-L25

If we created a  MavenParser we could DRY the code up.  Really the bin/parse_pom_xml logic could eventually be moved into that class.  If it memoized (we would need a mutex to keep our concurrent code happy) the result we could also avoid parsing the same file 2x.  

Thoughts?

maitrayshah: Yup, it makes sense if we moved it to its own class, currently just followed the existing structure but yes makes sense. This will become even more apparent when we do a gradle parser because it takes a long time so having one class return data to both CVE Scanners and Reporting Scanners would be nice.

bin/parse_pom_xml
```diff
@@ -22,17 +22,28 @@ if parsed_file.errors.size.positive?
   exit(1)
 end
 
-dependencies = []
+properties = {}
```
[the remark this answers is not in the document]
maitrayshah: Yes, need documentation for OSV as a whole and both GoOSV and MavenOSV.

lib/salus/scanners/osv/maven_osv.rb
```diff
@@ -0,0 +1,152 @@
...
+                end
+              elsif version_ranges["events"].length == 2
+                if version_ranges["events"][0].key?("introduced") &&
+                    version_ranges["events"][1].key?("fixed")
+                  introduced = version_ranges["events"][0]["introduced"]
+                  fixed = version_ranges["events"][1]["fixed"]
+                end
+              end
+
+              if version_ranges["type"] == "SEMVER" || version_ranges["type"] == "ECOSYSTEM"
```
joshua-ostrom, line 106: minor nit
```
if ["SEMVER", "ECOSYSTEM"].include?(version_ranges["type"])
```

maitrayshah: Good catch.

lib/salus/scanners/osv/maven_osv.rb
```diff
@@ -0,0 +1,152 @@
...
+                if version_matching(version, introduced, fixed)
+                  results.append({
```
joshua-ostrom, line 108: would you mind including a comment w/a sample "m" variable dump.  That help make clear what we are using vs leaving on the cutting room floor here (if any)

lib/salus/scanners/osv/maven_osv.rb
```diff
@@ -0,0 +1,152 @@
...
+      if @osv_vulnerabilities.nil?
+        err_msg = "MavenOSV: No vulnerabilities found to compare."
+        report_stderr(err_msg)
+        report_error(err_msg)
+        return
+      end
+
+      # Fetch vulnerable dependencies.
+      # Dedupe and select Github Advisory over other sources if available.
+      results = []
```
saikat056, line 37: It would be great if we can extract a method out of this.

```
results = vulnerable_dependencies_for(dependencies)
....

private
def vulnerable_dependencies_for(dependencies)
  results = []
  grouped = match_vulnerable_dependencies(dependencies).group_by { |d| d[:ID] }
  ...
  results
end
```

lib/salus/scanners/osv/maven_osv.rb
```diff
@@ -0,0 +1,152 @@
+require 'salus/scanners/osv/base'
+
+module Salus::Scanners::OSV
+  class MavenOSV < Base
+    class SemVersion < Gem::Version; end
+
```
saikat056, line 6: Please move the `private` methods under `private` section.

lib/salus/scanners/osv/maven_osv.rb
maitrayshah: Yup makes sense, let me update that.

lib/salus/scanners/osv/maven_osv.rb
```diff
@@ -0,0 +1,152 @@
...
+            bugsnag_notify("MavenOSV: Found #{lib}:#{version} with incompatible format.")
+            next
+          end
+
+          package_matches = @osv_vulnerabilities.select do |v|
+            v.dig("package", "name") == lib
+          end
+          package_matches.each do |m|
+            m["ranges"].each do |version_ranges|
+              introduced = fixed = ""
```
saikat056, line 93:
```
introduced, fixed = vulnerability_info_for(version_range) 
....

private
    
def vulnerability_info_for(version_range)      
      introduced = version_range["events"]&.first&.[]("introduced")
      fixed = version_range["events"]&.[](1)&.[]("fixed")

      [introduced.nil? "" : introduced, fixed.nil? "" : fixed]
end
```

lib/salus/scanners/osv/maven_osv.rb
```diff
@@ -0,0 +1,152 @@
...
+      vulnerable_flag
+    end
+
+    # Compare vulnerabilities found with dependencies found
+    # and return vulnerable dependencies
+    def match_vulnerable_dependencies(dependencies)
+      results = []
+      dependencies.each do |dependency|
+        lib = "#{dependency.fetch('group_id', '')}:#{dependency.fetch('artifact_id', '')}"
+        unless dependency['version'].nil?
```
saikat056, line 80: `unless dependency['version'].nil?` -> `if dependency['version']`

Please avoid double negative. Martin fowler has good article on this: https://refactoring.com/catalog/removeDoubleNegative.html

lib/salus/scanners/osv/maven_osv.rb
saikat056, line 106: `update_vul_details_for(results, package_match, introduced, fixed, version_range)`

```
    def update_vul_details_for(results, package_match, introduced, fixed, version_range)
      return unless valid_varsion_range?(version_range)
      return unless version_matching(version, introduced, fixed)

      results.append({
        "Package": package_match.dig("package", "name"),
        "Vulnerable Version": introduced,
        "Version Detected": version,
        "Patched Version": fixed,
        "ID": package_match.fetch("aliases", [package_match.fetch("id", [])])[0],
        "Database": package_match.fetch("database"),
        "Summary": package_match.fetch("summary", package_match.dig("details")).strip,
        "References": package_match.fetch("references", []).collect do |p|
          p["url"]
        end.join(", "),
        "Source":  package_match.dig("database_specific", "url") || DEFAULT_SOURCE,
        "Severity": package_match.dig("database_specific", "severity") || DEFAULT_SEVERITY
      })
    end

    def valid_varsion_range?(version_range)
      version_range["type"] == "SEMVER" || version_range["type"] == "ECOSYSTEM"
    end
```

lib/salus/scanners/osv/maven_osv.rb
```diff
@@ -0,0 +1,152 @@
...
+          end
+        end
+      end
+      results
+    end
+
+    # Find dependencies from the project
+    def find_dependencies
+      dependencies = []
+      shell_return = run_shell("bin/parse_pom_xml #{@repository.pom_xml_path}", chdir: nil)
```
saikat056, line 134: Place a new line here please

lib/salus/scanners/osv/maven_osv.rb
maitrayshah: Made changes to it so should be much more readable now.

lib/salus/package_utils/maven_dependency_parser.rb
```diff
@@ -0,0 +1,60 @@
...
+        version = properties.dig(version.children.to_s) || version.children.to_s unless version.nil?
+
+        parsed.append({
+                        "group_id" =>  group_id.nil? ? nil : group_id.children.to_s,
+                        "artifact_id" => artifact_id.nil? ? nil : artifact_id.children.to_s,
+                        "version" => version
+                      })
+      end
+
+      @pom_xml_dependencies = parsed
```
saikat056, line 57:
```ruby
def parsed_dependencies
      parsed = []
      @pom_xml_content.css('dependency')&.each do |dependency|
        group_id = dependency.at('groupId')
        artifact_id = dependency.at('artifactId')
        version = dependency.at('version')
        version = properties.dig(version.children.to_s) || version.children.to_s unless version.nil?

        parsed.append({
                        "group_id" =>  group_id.nil? ? nil : group_id.children.to_s,
                        "artifact_id" => artifact_id.nil? ? nil : artifact_id.children.to_s,
                        "version" => version
                      })
      end
   
      @pom_xml_dependencies = parsed
end
```

spec/fixtures/npm_audit/success_with_exceptions/salus-sarif.yaml
```diff
@@ -0,0 +1,13 @@
...
+            - {
+                advisory_id: "1048867",
+                changed_by: "ryan.sears",
+                notes: "See https://www.npmjs.com/advisories/39. We're not vulnerable to this because uglify is a transient dev dependency, and we never actually invoke it with our code.",
+              }
+            - {
+                advisory_id: "1048868",
+                changed_by: "ryan.sears",
+                notes: "See https://www.npmjs.com/advisories/48. We're not vulnerable to this because this is a regex dos and we have nothing that puts user input into it. The impact is also minimal.",
+              }
```
saikat056, line 13: Please add a new line to satisfy Github

spec/fixtures/npm_audit/success_with_exceptions/salus.yaml
```diff
@@ -10,4 +10,4 @@ scanner_configs:
                 advisory_id: "1006709",
                 changed_by: "ryan.sears",
                 notes: "See https://www.npmjs.com/advisories/48. We're not vulnerable to this because this is a regex dos and we have nothing that puts user input into it. The impact is also minimal.",
-              }
+              }
```
saikat056, line 13: Please add a new line to satisfy Github

spec/fixtures/osv/maven_osv/success_no_vulnerability/pom.xml
```diff
@@ -0,0 +1,22 @@
...
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.mybatis</groupId>
+            <artifactId>mybatis</artifactId>
+            <version>${mybatis.version}</version>
+        </dependency>
+    </dependencies>
+</project>
```
saikat056, line 22: Please add a new line to satisfy Github

spec/fixtures/osv/maven_osv/failure_vulnerability_present/pom.xml
```diff
@@ -0,0 +1,113 @@
...
+                <configuration>
+
+                    <failOnMissingWebXml>false</failOnMissingWebXml>
+                </configuration>
+            </plugin>
+        </plugins>
+
+        <finalName>Example</finalName>
+    </build>
+</project>
```
saikat056, line 113: Please add a new line to satisfy Github

ghbren (approved):

lib/salus/scanners/report_pom_xml.rb
```diff
@@ -7,24 +7,16 @@ class ReportPomXml < Base
...
-      end
-
-      dependencies = nil
       begin
-        dependencies = JSON.parse(shell_return.stdout)
-      rescue JSON::ParserError
-        err_msg = "Could not parse JSON returned by bin/parse_pom_xml's stdout!"
-        report_stderr(err_msg)
-        report_error(err_msg)
+        parser = Salus::MavenDependencyParser.new(@repository.pom_xml_path)
```
joshua-ostrom, line 11: 👍

joshua-ostrom (approved):
Thanks for making the changes!

PR #567, opened 2022-03-14
Updating Package Version Scanner

maitrayshah:
- Update handling of `min_version` and `max_version` to work independently.
- Add support for `block` to handle blocking specific versions of a package.
- Add extra checks if config missing / not correct format.
- Fix for correctly handling `pass_on_raise: True`


lib/salus/scanners/package_version/npm_package_version_scanner.rb
```diff
@@ -11,32 +11,58 @@ def should_run?
...
+        line_number = @dependencies[package_name][repo_version.to_s].to_s
+        error_msg = "Package version for (#{package_name}) (#{repo_version})" \
+        "is less than minimum configured version (#{min_version}) on line " \
+        "{#{line_number}} in package-lock.json"
+        report_error(error_msg)
+        @passed = false
+      end
+    end
+
+    def check_max_version(package_name, repo_version, max_version)
```
ghbren, line 42: Maybe refactor `check_min_version` and `check_max_version`, because they share some code?

ghbren: If I remember correctly, you also need to `report_stderr`, because `report_error` doesn't display anything to the txt output, it only displays to the json.

lib/salus/scanners/package_version/npm_package_version_scanner.rb
```diff
@@ -11,32 +11,58 @@ def should_run?
...
-        if repo_version && repo_version > max_version
+    def check_blocked_versions(package_name, repo_version, blocked_versions)
+      blocked_versions.each do |blocked|
+        if repo_version == blocked
           line_number = @dependencies[package_name][repo_version.to_s].to_s
           error_msg = "Package version for (#{package_name}) (#{repo_version}) " \
-          "is greater than maximum configured version (#{max_version}) on line "\
+          "matches the configured blocked version (#{blocked}) on line "\
           "{#{line_number}} in package-lock.json"
           report_error(error_msg)
```
ghbren, line 60: `report_stderr` too?

joshua-ostrom (approved):

PR #574, opened 2022-03-25
upgrade go to 1.18

quentinlesceller:
As titled.


joshua-ostrom:
Thank you!  We captured golang / gosec upgrades in https://github.com/coinbase/salus/pull/554

PR #585, opened 2022-04-14
Updating ReportNodeModules to support latest yarn

maitrayshah:
- Adding support for latest yarn version in ReporNodeModules.


lib/salus/scanners/report_node_modules.rb
```diff
@@ -58,7 +62,72 @@ def self.supported_languages
...
+      else
+        record_dependencies_from_legacy_yarn_lock
+      end
+    end
+
+    def record_dependencies_from_latest_yarn_lock
+      packages_shell_return = run_shell(
+        YARN_LIST_PACKAGES_COMMAND, chdir: @repository.path_to_repo
+      )
+      unless packages_shell_return.success?
```
ghbren, line 83: If I'm not mistaken, `run_shell.success?` is based on exit code.  
Not all commands return the standard success/failure exit codes.
Did you check whether yarn follows the the standard codes?

lib/salus/scanners/report_node_modules.rb
```diff
@@ -58,7 +62,72 @@ def self.supported_languages
...
+        raise ParseError, "Failed to parse yarn.lock file: #{packages_shell_return.stderr}"
```
ghbren, line 84: Is it possible for the command to return errors unrelated to parsing yarn.lock?
If so, maybe update the error message?

lib/salus/scanners/report_node_modules.rb
```diff
@@ -58,7 +62,72 @@ def self.supported_languages
...
+      end
+
+      source_shell_return = run_shell(
+        YARN_REGISTRY_SOURCE_COMMAND, chdir: @repository.path_to_repo
+      )
+      unless source_shell_return.success?
+        raise ParseError, "Failed to fetch npmRegistryServer: #{source_shell_return.stderr}"
```
ghbren, line 91: Not a blocker, but this is not exactly a ParseError?

lib/salus/scanners/report_node_modules.rb
```diff
@@ -58,7 +62,72 @@ def self.supported_languages
...
+      dependencies = []
+      # source shell output - https://registry.yarnpkg.com
+      source = source_shell_return.stdout.strip
+
+      yarn_packages = packages_shell_return.stdout
+      # yarn info shell output -
+      # "gulp-cssmin@npm:0.2.0"
+      # "gulp-modify-file@npm:1.0.1"
+      yarn_packages.each_line do |line|
+        name = line.split("@npm")[0]
```
ghbren, line 103: is it guaranteed that each line has `@npm` in the middle??

ghbren: or would it be better to split by `@` (if it exists)?

lib/salus/scanners/report_node_modules.rb
```diff
@@ -58,7 +62,72 @@ def self.supported_languages
...
+        version = line.split("@npm")[1]
+        if name.present? && version.present?
+          # Handle patched packages. Sample format:
+          # "node-fetch-web-streams@patch:node-fetch-web-streams@npm%3A1.0.1#.yarn/patches/
+          # node-fetch-web-streams-npm-1.0.1-63d9cb5d48.patch::version=1.0.1&hash=804812&locator=%40
+          # frontend%2Fmobile%40workspace%3A."
+          version = version[/\::version=(.*?)&/, 1] if version.include?("::version=")
```
ghbren, line 110: What is `(.*?)` supposed to mean?  
It seems to mean 0 or 1 of "0 or more of any character", which seems a little confusing to me.

lib/salus/scanners/report_node_modules.rb
maitrayshah: Could be haven't seen it yet. Yes let update the error message.

lib/salus/scanners/report_node_modules.rb
maitrayshah: Good catching, checking the documentation it seems like it follows this - https://yarnpkg.com/advanced/error-codes

lib/salus/scanners/report_node_modules.rb
maitrayshah: Mainly to match non greedy, since I only want the first match.

ghbren (dismissed):

ghbren (approved):

PR #588, opened 2022-04-20
add logs for misconfigured ignore value

Kamsiy:
## Description
Certain Scanner configurations for brakeman causes salus to raise errors. This PR handles these errors and also adds a log to buildkite
## Issue 1
the `ignore` configuration option was given an Array instead of a string
```
scanner_configs:
  Brakeman:
    ignore: 
    - config/brakeman.ignore
```
### Currently 
Screen Shot 2022-04-20 at 12 01 04 PM
### Solution
Screen Shot 2022-04-20 at 11 55 02 AM

## Issue 2
the `ignore` configuration option is given a path that does not exist
```
scanner_configs:
  Brakeman:
    ignore: brakeman.ignore
```
### Currently
Screen Shot 2022-04-20 at 11 59 57 AM

### Solution
Screen Shot 2022-04-20 at 11 56 02 AM


lib/salus/scanners/brakeman.rb
```diff
@@ -133,6 +133,9 @@ def user_supplied_ignore_expirations?
         return true if item.key?('expiration')
       end
       false
+    rescue TypeError, Errno::ENOENT => e
+      report_error(e.message)
+      bugsnag_notify(e)
```
ghbren, line 138: Please delete `bugsnag_notify`, because `report_error` already calls `bugsnag_notify`.
Ref: https://github.com/coinbase/salus/blob/e66332c8479dd4b220cc22af9f588abc8848dfc0/lib/salus/scanners/base.rb#L188-L194

lib/salus/scanners/brakeman.rb
```diff
@@ -133,6 +133,9 @@ def user_supplied_ignore_expirations?
         return true if item.key?('expiration')
       end
       false
+    rescue TypeError, Errno::ENOENT => e
```
ghbren, line 136: Could you rescue `StandardError` instead, since it's a super class of `TypeError`?
That way, we could potentially catch more errors.

ghbren: Also, I'm not sure why the `rescue` includes `Error::ENOENT`, which means no file or directory.  
There is nothing in this function that could possibly throw this type of exception.

lib/salus/scanners/brakeman.rb
```diff
@@ -110,6 +110,18 @@ def merged_ignore_file_contents
...
 
+      if !@config['ignore'].is_a?(String)
+        msg = "Brakeman ignore path #{@config['ignore']} is a/an #{@config['ignore'].class} and"\
+        " not a String. "
+        report_error(msg)
+        return []
+      end
+
+      if !File.exist?(@config['ignore'])
+        msg = "Brakeman ignore path #{@config['ignore']} does not exist"
```
ghbren, line 121: Please test by build local container and run salus.  
Try to use a valid brakeman ignore path and the code will still report the error.

ghbren: You probably need a `File.join` like here https://github.com/coinbase/salus/blob/2f3dbcba104c83dcd8dee178b97b8075d3c44e6f/lib/salus/scanners/brakeman.rb#L235

ghbren: It would be great to add a test case too.

PR #589, opened 2022-04-21
Breakout SARIF Active Scanners from Suppressed

joshua-ostrom:
Currently Salus is overloading SARIF suppressed results to apply for both results that are supressed due to exceptions on file as well as findings from active (not enforced scanners). 

This PR provides a new configuration option for SARIF reporting:
"include_active"

When true the SARIF will contain findings from active scanners.  By default the SARIF will ~*not*~ contain findings from unenforced scanners.

Props to https://github.com/geraldnash-cb/salus/commit/8acffabf442f03875fcbb5f2153515084e45f897 for standing up initial support - opening a branch on the main repo to enable wider team collaboration.  Changes added from the referenced commit include a few additional spec checks and the removal of suppression info for active only scanners


lib/sarif/base_sarif.rb
```diff
@@ -134,16 +134,18 @@ def build_runs_object(supported)
 
         next if !parsed_issue
 
-        next if parsed_issue[:suppressed] && @config['include_suppressed'] == false
+        not_required = (@required.nil? || @required == false)
 
-        next if @required == false && @config['include_suppressed'] == false
+        next if not_required && @config.fetch('include_suppressed', true) == false
+
+        next if not_required && @config.fetch('include_active', true) == false
```
ghbren, line 141: It would be better to combine the two lines above into something like the following, to avoid duplicated code

```ruby
x = (!@config.fetch('include_suppressed', true)) ||| !(@config.fetch('include_active', true))
next if not_required && x
```

docs/configuration.md
```diff
@@ -80,6 +80,8 @@ cascade_config:
 # The available options for the sarif_options keyword are:
 # 1) `include_suppressed: true/false` -This option allows users to include/exclude suppressed/excluded results 
 #    in their sarif reports. Currently this is supported for NPM audit reports
+# 2) `include_active: true/false` - This option allows users to include or exclude results from active (non enforced)
+#    scanners in their sarif reports.  Default is false.
```
ghbren, line 84: I wonder if we should rename `include_active` to `include_non_enforced`?

All scanners that run in a salus scan are active scanners.

ghbren: Our goal here is to exclude non-enforced active scanners, not active scanners.

maitrayshah: +1

ghbren:
Bug found. The doc says below, which is not the actual behavior.
> include_active: true/false` - This option allows users to include or exclude results from active (non enforced)
>    scanners in their sarif reports.  Default is false.

Try to run salus from the salus repo with this salus.yaml
```yaml
reports:
  - uri: file://out.json
    format: sarif

enforced_scanners:
  - RepoNotEmpty

reports:
  - uri: file://out.json
    format: sarif
    sarif_options:
      include_active: false
```

Take a look at the sarif (out.json), it does not contain Gosec results, which is fine.

Now, edit salus.yaml to remove `include_active:false`. The new sarif contains Gosec results, but it shouldn't, because `include_active` should default to `false`, based on the doc.

docs/configuration.md
joshua-ostrom: Sounds good will rename

joshua-ostrom:
> Bug found. The doc says below, which is not the actual behavior.
> 
> > include_active: true/false` - This option allows users to include or exclude results from active (non enforced)
> > scanners in their sarif reports.  Default is false.
> 
> Try to run salus from the salus repo with this salus.yaml
> 
> ```yaml
> reports:
>   - uri: file://out.json
>     format: sarif
> 
> enforced_scanners:
>   - RepoNotEmpty
> 
> reports:
>   - uri: file://out.json
>     format: sarif
>     sarif_options:
>       include_active: false
> ```
> 
> Take a look at the sarif (out.json), it does not contain Gosec results, which is fine.
> 
> Now, edit salus.yaml to remove `include_active:false`. The new sarif contains Gosec results, but it shouldn't, because `include_active` should default to `false`, based on the doc.

Thanks Bri, I'll update the doc.... it's out of date we initially had it defaulting to false but aligned to defaulting to true

maitrayshah:
For the following config - 

```
reports:
  - uri: file://salus-report.sarif
    format: sarif
    sarif_options:
      include_active: false
```

Where there is no enforced and we will run all scanners that applies as active scanners. The expected outcome should be empty SARIF ?

ghbren:
> For the following config -
> 
> ```
> reports:
>   - uri: file://salus-report.sarif
>     format: sarif
>     sarif_options:
>       include_active: false
> ```
> 
> Where there is no enforced and we will run all scanners that applies as active scanners. The expected outcome should be empty SARIF ?

I haven't tried this, but the SARIF should not be empty, it would contain something like this for every scanner

```json
      "tool": {
        "driver": {
          "informationUri": "https://github.com/securego/gosec",
          "name": "Gosec",
          "properties": {
            "salusEnforced": false
          },
          "rules": [

          ],
          "version": "2.11.0"
        }
      }
    },
```

ghbren (approved):

PR #593, opened 2022-04-26
Yarn3 audit updates

maitrayshah:
- Adding --recursive: This will enable alerting on transitive dependencies. 
- Adding option to ignore `devDependencies`.
```
  YarnAudit:
    exclude_groups:
      - devDependencies
```

- Updating documentation.
- Adding new config var `'scan_depth'` for latest versions of Yarn.


lib/salus/scanners/yarn_audit.rb
```diff
@@ -28,7 +30,15 @@ def run
 
     def handle_latest_yarn_audit
       vulns = []
-      shell_return = run_shell(LATEST_YARN_AUDIT_COMMAND)
+
+      dep_types = @config.fetch('exclude_groups', [])
+      audit_command = if dep_types.include?('devDependencies')
```
joshua-ostrom, line 35: Can you update the docs at https://github.com/coinbase/salus/blob/master/docs/scanners/yarn_audit.md to convey the nuance w/what yarn 3 supports?  i.e. optional dependencies etc won't apply for yarn 3

lib/salus/scanners/yarn_audit.rb
```diff
@@ -10,7 +10,9 @@ class YarnAudit < NodeAudit
     # the command was previously 'yarn audit --json', which had memory allocation issues
     # see https://github.com/yarnpkg/yarn/issues/7404
     LEGACY_YARN_AUDIT_COMMAND = 'yarn audit --no-color'.freeze
-    LATEST_YARN_AUDIT_COMMAND = 'yarn npm audit --all --json'.freeze
+    LATEST_YARN_AUDIT_ALL_COMMAND = 'yarn npm audit --recursive --all --json'.freeze
```
joshua-ostrom, line 13: Can we add a config option for recursive?

Does yarn <3 support recursive?  If so we should mirror behavior between yarns (ideally both using the same config var to determine if the scan will be recursive)

geraldnash-cb (approved):

PR #595, opened 2022-04-28
Support Gradle version 6 and 7

Kamsiy:
This PR 
- adds support for gradle version 6 and 7
- projects using a gradle version not supported by Salus are shown upgrade notifications. 
Screen Shot 2022-04-28 at 12 56 21 PM


spec/lib/salus/scanners/osv/gradle_osv_spec.rb
```diff
@@ -137,5 +137,34 @@ def stub_req_with_valid_response
...
+      let(:path_v7) { "version_7" }
+      let(:fixture_path) { File.expand_path(path_str, __dir__) }
+
+      it 'runs gradle version 7 successfully' do
+        repo = Salus::Repo.new(File.join(fixture_path, path_v7))
+        scanner = Salus::Scanners::OSV::GradleOSV.new(repository: repo, config: {})
+        stub_req_with_valid_response
+        scanner.run
+        expect(scanner.report.to_h.fetch(:passed)).to eq(true)
+      end
```
saikat056, line 153: Please put a new line after each it do..end

Dockerfile
```diff
@@ -28,12 +28,17 @@ RUN apt-get update && apt-get upgrade -y --no-install-recommends && apt-get inst
 WORKDIR /root
 
 ### JDK
-RUN wget https://download.java.net/java/GA/jdk17.0.2/dfd4a8d0985749f896bed50d7138ee7f/8/GPL/openjdk-17.0.2_linux-x64_bin.tar.gz -P /tmp
-RUN tar xvf /tmp/openjdk-17.0.2_linux-x64_bin.tar.gz -C /
+RUN wget https://download.java.net/java/GA/jdk11/9/GPL/openjdk-11.0.2_linux-x64_bin.tar.gz -P /tmp
+RUN tar xvf /tmp/openjdk-11.0.2_linux-x64_bin.tar.gz -C /
```
saikat056, line 32: We are downgrading jdk version from 17.0.2 to 11.0.2. Will there be any security concern? 🤔

Kamsiy: Java 17 is supported by Gradle version 7. We could bump this up to java 13 since `gradle 6` is compatible with it.

Here is a link to the gradle and jdk compatibility matrix. https://docs.gradle.org/current/userguide/compatibility.html

Please let me know if you want this bumped

joshua-ostrom: jdk 11 is supported till 2026 - that should give time for projects using older gradles to upgrade and we can then pull support for the older version of java / gradles

lib/salus/package_utils/gradle_dependency_parser.rb
```diff
@@ -0,0 +1,30 @@
...
+    if !shell_result.success?
+      report_error("Gradle Version Not supported. Please Upgrade to gradle version 6 and above")
+      return []
+    end
+
+    dependencies = []
+
+    shell_result.stdout.scan(dependency_metadata_regex).each do |dependency_properties|
+      dependency_hash = {}
+      dependency_hash['group_id'] = dependency_properties[0]
```
joshua-ostrom, line 21: We should probably report an error here if the dependency_properties.size < 3

joshua-ostrom (approved):

PR #596, opened 2022-05-05
Eliminate bundle audit vulnerabilities that are likely not introduced by the commit.

ghbren:
Code stores the gem name without version in snippet.  Then looks for the snippet in git diff.
If there are multiple versions of the same gem, then it may be an issue. However, this PR is just here to eliminate vulnerabilities that are not likely introduced commit.

PR assumes Gemfile.lock was auto-created.  If the user manually updates it to have some unconventional format, then things may break.

PR has also been tested using docker build ..., docker run local salus with `--sarif_full_dif ... --git-diff ...` to make sure
1.  vuls caused by changed lines do not show up in the diffed results.
2. vuls introduced by the commit still show up.

### DIFF
--- a/lib/sarif/bundle_audit_sarif.rb
+++ b/lib/sarif/bundle_audit_sarif.rb
@@ -29,7 +29,7 @@ module Sarif
       if issue[:line_number]
         result[:start_line] = issue[:line_number]
         result[:start_column] = 1
-        result[:code] = issue[:name].to_s
+        result[:code] = issue[:name].to_s # Code stores the gem name without version in snippet
       end
 
       if issue[:type] == "InsecureSource"
--- a/lib/sarif/bundle_audit_sarif.rb
+++ b/lib/sarif/bundle_audit_sarif.rb
@@ -29,6 +29,7 @@ module Sarif
       if issue[:line_number]
         result[:start_line] = issue[:line_number]
         result[:start_column] = 1
+        result[:code] = issue[:name].to_s
       end
 
       if issue[:type] == "InsecureSource"
@@ -49,6 +50,13 @@ module Sarif
       when 7.0..10.0
         SARIF_WARNINGS[:error]
       end
+    end
+
+    def self.snippet_possibly_in_git_diff?(snippet, lines_added)
+      # snippet in sarif just has the package name (without spaces/version)
+      # actual snippet in Gemfile.lock looks like " nokogiri (>= 1.5.11, < 2.0.0)"
+      snippet = ' ' + snippet + ' ('
+      lines_added.keys.any? { |line| line.include? snippet }
     end
   end
 end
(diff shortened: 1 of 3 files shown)


lib/sarif/bundle_audit_sarif.rb
```diff
@@ -50,5 +51,12 @@ def sarif_level(severity)
         SARIF_WARNINGS[:error]
       end
     end
+
+    def self.snippet_possibly_in_git_diff?(snippet, lines_added)
```
joshua-ostrom, line 55: Is the only used by specs? If so can we make this an rspe helper?

lib/sarif/bundle_audit_sarif.rb
```diff
@@ -29,6 +29,7 @@ def parse_issue(issue)
       if issue[:line_number]
         result[:start_line] = issue[:line_number]
         result[:start_column] = 1
+        result[:code] = issue[:name].to_s
```
joshua-ostrom, line 32: Can you include the comment present in your PR description here in the code for readability?
# Code stores the gem name without version in snippet.

lib/sarif/bundle_audit_sarif.rb
ghbren: No. This code is used outside specs.
In fact, `docker run ... --git-diff` calls this method.

lib/sarif/bundle_audit_sarif.rb
ghbren: added, great suggestion!

joshua-ostrom (approved):

spec/lib/sarif/bundle_audit_spec.rb
```diff
@@ -135,4 +138,30 @@
...
+        snippet = 'helloworld'
+        git_diff = File.read(git_diff_file)
+        new_lines_in_git_diff = Sarif::BaseSarif.new_lines_in_git_diff(git_diff)
+        r = Sarif::BundleAuditSarif.snippet_possibly_in_git_diff?(snippet, new_lines_in_git_diff)
+        expect(r).to be true
+        snippet = 'bye'
+        r = Sarif::BundleAuditSarif.snippet_possibly_in_git_diff?(snippet, new_lines_in_git_diff)
+        expect(r).to be true
+
+        # should not match part of the package name
```
saikat056, line 155: Can we write this part as a separate unit test?

```
it 'should not match part of the package name' do
  ...
end
```

spec/lib/sarif/bundle_audit_spec.rb
```diff
@@ -135,4 +138,30 @@
...
+        snippet = 'hello'
+        r = Sarif::BundleAuditSarif.snippet_possibly_in_git_diff?(snippet, new_lines_in_git_diff)
+        expect(r).to be false
+
+        # should not match package that was in git diff but not added with this commit
```
saikat056, line 160: Please write this as a separate unit test

it 'should not match package that was in git diff but not added with this commit' do
end

saikat056 (approved):

PR #597, opened 2022-05-06
combine yarn vulnerabilities with the same id

ghbren:


lib/salus/scanners/yarn_audit.rb
```diff
@@ -184,6 +186,26 @@ def normalize_vuln(vuln)
...
+          uniq_vulns[id]['Dependency of'].push vul['Dependency of']
+        else
+          uniq_vulns[id] = vul
+          uniq_vulns[id]['Dependency of'] = [uniq_vulns[id]['Dependency of']]
+        end
+      end
+
+      vulns = uniq_vulns.values
+      vulns.each do |vul|
+        vul['Dependency of'] = vul['Dependency of'].sort.join(', ')
```
joshua-ostrom, line 204: Can we have any cases where we may have the same "dependency of" listed more than once?  I assume we won't.  If we will we may want a .uniq.sort here.

spec/lib/salus/scanners/yarn_audit_spec.rb
```diff
@@ -71,7 +71,7 @@
 
       expect(scanner.report.to_h.fetch(:passed)).to eq(false)
       vulns = JSON.parse(scanner.report.to_h[:info][:stdout]).sort { |a, b| a["ID"] <=> b["ID"] }
-      expect(vulns.size).to eq(20)
+      expect(vulns.size).to eq(17)
```
joshua-ostrom, line 74: 👍 Lowering noise for the user is a win

joshua-ostrom (approved):

PR #607, opened 2022-05-16
[SPIKE] Add Slither Scanner to Salus

geraldnash-cb:
Adding [Slither](https://github.com/crytic/slither) to Salus.

Note these changes do _not_ include a SARIF adapter for the scanner.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -198,6 +198,7 @@ RUN ln -sf /usr/local/go/bin/go /usr/local/bin
 RUN python -m easy_install pip==${PIP_VERSION} \
   && python3 -m easy_install pip==${PIP_VERSION}
 
+# Install solidity compiler
 ENV SOLC_VERSION 0.8.4
 RUN solc-select install ${SOLC_VERSION} && solc-select use ${SOLC_VERSION}
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -68,6 +68,16 @@ RUN pip install wheel \
   && mv .local/bin/bandit .local/bin/bandit2 \
   && pip3 install --user bandit==${BANDIT_VERSION}
 
+# Install solidity compiler version manager
+ENV SOLC_SELECT_VERSION 0.2.1
+
+RUN pip3 install --user solc-select==${SOLC_SELECT_VERSION}
+ENV PATH="/root/.local/bin:${PATH}"
+
+# Install slither, solidity static code scanner
+ENV SLITHER_VERSION 0.8.3
+RUN pip3 install --user --no-cache-dir slither-analyzer==${SLITHER_VERSION}
+
 
 ### Ruby
 # ruby gems
@@ -187,6 +197,9 @@ COPY --from=builder /opt/gradle/gradle-6.9.2 /opt/gradle/gradle-6.9.2
 RUN ln -sf /usr/local/go/bin/go /usr/local/bin
 RUN python -m easy_install pip==${PIP_VERSION} \
   && python3 -m easy_install pip==${PIP_VERSION}
+
+ENV SOLC_VERSION 0.8.4
+RUN solc-select install ${SOLC_VERSION} && solc-select use ${SOLC_VERSION}
 
 ### Salus
 WORKDIR /home
--- a/Dockerfile
+++ b/Dockerfile
@@ -68,15 +68,14 @@ RUN pip install wheel \
   && mv .local/bin/bandit .local/bin/bandit2 \
   && pip3 install --user bandit==${BANDIT_VERSION}
 
-# Install solidity compiler version manager
+# Install solidity compiler version manager and 
+# slither, a solidity static code scanner
 ENV SOLC_SELECT_VERSION 0.2.1
-
-RUN pip3 install --user solc-select==${SOLC_SELECT_VERSION}
-ENV PATH="/root/.local/bin:${PATH}"
-
-# Install slither, solidity static code scanner
 ENV SLITHER_VERSION 0.8.3
-RUN pip3 install --user --no-cache-dir slither-analyzer==${SLITHER_VERSION}
+
+RUN pip3 install --user solc-select==${SOLC_SELECT_VERSION} \
+  && pip3 install --user --no-cache-dir slither-analyzer==${SLITHER_VERSION}
+ENV PATH="/root/.local/bin:${PATH}"
 
 
 ### Ruby
--- a/lib/salus/repo.rb
+++ b/lib/salus/repo.rb
@@ -39,7 +39,9 @@ module Salus
       # Swift
       { handle: :package_resolved, filename: 'Package.resolved' },
       # Apple Ecosystem (macOS, iOS, etc)
-      { handle: :podfile_lock, filename: 'Podfile.lock' }
+      { handle: :podfile_lock, filename: 'Podfile.lock' },
+      # Solidity
+      { handle: :sol_file, filename: '.sol', wildcard: true }
     ].freeze
 
     # Define file checkers.
--- a/lib/salus/repo.rb
+++ b/lib/salus/repo.rb
@@ -41,7 +41,11 @@ module Salus
       # Apple Ecosystem (macOS, iOS, etc)
       { handle: :podfile_lock, filename: 'Podfile.lock' },
       # Solidity
-      { handle: :sol_file, filename: '.sol', wildcard: true }
+      { handle: :sol_file, filename: '.sol', wildcard: true },
+      { handle: :truffle_js, filename: 'truffle.js' },
+      { handle: :truffle_ts, filename: 'truffle.ts' },
+      { handle: :hardhat_config_js, filename: 'hardhat.config.js' },
+      { handle: :hardhat_config_ts, filename: 'hardhat.config.ts' }
     ].freeze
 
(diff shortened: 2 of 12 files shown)


Dockerfile
```diff
@@ -68,6 +68,16 @@ RUN pip install wheel \
...
 
+# Install solidity compiler version manager
+ENV SOLC_SELECT_VERSION 0.2.1
+
+RUN pip3 install --user solc-select==${SOLC_SELECT_VERSION}
+ENV PATH="/root/.local/bin:${PATH}"
+
+# Install slither, solidity static code scanner
+ENV SLITHER_VERSION 0.8.3
+RUN pip3 install --user --no-cache-dir slither-analyzer==${SLITHER_VERSION}
```
joshua-ostrom, line 79: Can you combine these pip3 install commands (that will result in a smaller image)?

lib/salus/config.rb
```diff
@@ -203,11 +203,12 @@ def fetch_envars(config_hash)
...
     def all_none_some(superset, subset)
-      if subset == 'all'
+      case subset
+      when 'all'
         Set.new(superset) # entire set
-      elsif subset == 'none'
+      when 'none'
         Set.new           # empty set
-      elsif subset.is_a?(Array)
+      when Array
```
saikat056, line 211: This part is not exactly same.

```
➜  Documents irb
irb(main):001:0> s = [2,3,4]
=> [2, 3, 4]
irb(main):002:0> s == Array
=> false
irb(main):003:0> s.is_a? Array
=> true
```

lib/salus/repo_searcher.rb
```diff
@@ -157,9 +157,8 @@ def run_rg(*args)
       data = IO.popen(args, chdir: @path_to_repo).read
       return [] if data == ""
 
-      files = data.lines.map(&:strip)
+      data.lines.map(&:strip)
       # files are all relative to @path_to_repo
```
saikat056, line 161: Can we remove the comment because we are not using **files** variable anymore?
# files are all relative to @path_to_repo

lib/salus/repo.rb
```diff
@@ -39,7 +39,9 @@ class Repo
       # Swift
       { handle: :package_resolved, filename: 'Package.resolved' },
       # Apple Ecosystem (macOS, iOS, etc)
-      { handle: :podfile_lock, filename: 'Podfile.lock' }
+      { handle: :podfile_lock, filename: 'Podfile.lock' },
+      # Solidity
+      { handle: :sol_file, filename: '.sol', wildcard: true }
```
ghbren, line 44: What does `wildcard: true` do?

lib/salus/scanners/slither.rb
```diff
@@ -0,0 +1,55 @@
...
+
+# Slither (https://github.com/crytic/slither) is a Python-based static
+# analysis tool for Solidity smart contracts
+
+module Salus::Scanners
+  class Slither < Base
+    include Salus::Formatting
+
+    def should_run?
+      !!@repository.sol_file_present?
```
ghbren, line 12: What does `!!` mean? I have never seen this syntax before.

lib/salus/scanners/slither.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+
+    def run
+      # Most Solidity projects will be Hardhat (https://hardhat.org/) or Truffle (https://trufflesuite.com/),
+      # which both use NPM to manage Solidity dependencies. We should NPM install in these cases to ensure
+      # Slither has the needed dependencies for scanning
+      run_shell("npm install", chdir: @repository.path_to_repo) if @repository.package_json_present?
```
ghbren, line 19: Does `npm install` follow the standard success/failure exit codes?

ghbren: Does `npm install` install dependencies to somewhere outside `@repository.path_to_repo`?

lib/salus/scanners/slither.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+
+    def self.supported_languages
+      ['solidity']
+    end
+
+    protected
+
+    def command
+      "slither . --json - --exclude-low --exclude-informational --exclude-optimization"
```
ghbren, line 52: This command is fine for the spike, but I feel we should support these excludes in salus.yaml.

lib/salus/scanners/slither.rb
```diff
@@ -0,0 +1,55 @@
...
+      !!@repository.sol_file_present?
+    end
+
+    def run
+      # Most Solidity projects will be Hardhat (https://hardhat.org/) or Truffle (https://trufflesuite.com/),
+      # which both use NPM to manage Solidity dependencies. We should NPM install in these cases to ensure
+      # Slither has the needed dependencies for scanning
+      run_shell("npm install", chdir: @repository.path_to_repo) if @repository.package_json_present?
+      shell_return = run_shell(command, chdir: @repository.path_to_repo)
+      stdout_json = JSON.parse(shell_return.stdout)
```
ghbren, line 21: json parse error should be checked here.

lib/salus/scanners/slither.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      return report_success if shell_return.success?
+
+      report_failure
```
ghbren, line 25: When vulnerabilities are found, it would help to print out a message on how to inline allowlist.  Maybe link the slither doc.

lib/salus/scanners/slither.rb
ghbren: Also, do we need to implement the exclude directory option?

lib/salus/scanners/slither.rb
geraldnash-cb: Because [sol_file uses wildcard matching](https://github.com/coinbase/salus/pull/607/files#diff-50c2ad55bc3977fdf09bc45a06d90d7febb43141bed71c7f8c49d503b44a90c8R44), `sol_file_present?` returns an array of `.sol` filepaths. `!!` lets us coerce the value of the array from truthy (non-empty array) or falsey (empty array) to `true` or `false`, respectively.

lib/salus/scanners/slither.rb
geraldnash-cb: I don't think so. Are you referring to excluding test directories? If so, I don't think we need to as the most popular frameworks (Truffle and Hardhat) have tests written in JavaScript. 

We should consider it for later implementation, though. [DappTools](https://github.com/dapphub/dapptools) and [Foundry](https://github.com/foundry-rs/foundry) let developers write tests in Solidity itself.

lib/salus/scanners/slither.rb
geraldnash-cb: [The `should_run?` tests](https://github.com/coinbase/salus/pull/607/files#diff-dadba0278118c6a057fb1d5f6ed9e58c5cce981b84578da134b4ce7dfff09068R5-R20) demonstrate this behavior.

lib/salus/scanners/slither.rb
```diff
@@ -0,0 +1,64 @@
...
+      'comment before the offending line.\n\nFor example, `//slither-disable-next-line timestamp` will disable '\
+      'detection of `block.timestamp` usage by disabling the `timestamp` detector.')
+        report_stdout(shell_return.stdout)
+        log(prettify_json_string(shell_return.stdout))
+      end
+    end
+
+    def version
+      shell_return = run_shell('slither --version')
+      shell_return.stdout.rstrip()
```
saikat056, line 51: Can we write this part as one line?

```
def version
   run_shell('slither --version').stdout.rstrip()
end
```

lib/salus/scanners/slither.rb
```diff
@@ -0,0 +1,64 @@
...
+          json_parser_error_msg = "Could not parse slither stdout"
+          "{ \"success\": false, \"error\": \"#{json_parser_error_msg}\" }"
+      end
+
+      return report_success if shell_return.success?
+
+      report_failure
+
+      if !stdout_json['success'] || !stdout_json['error'].nil?
+        # puts prettify_json_string(shell_return.stdout)
```
saikat056, line 35: Do we need this comment here? Looks like following line has been commented out for debugging purpose.

`# puts prettify_json_string(shell_return.stdout)`

lib/salus/scanners/slither.rb
```diff
@@ -0,0 +1,62 @@
...
+    def should_run?
+      !!@repository.sol_file_present?
+    end
+
+    def run
+      # Most Solidity projects will be Hardhat (https://hardhat.org/) or Truffle (https://trufflesuite.com/), 
+      # which both use NPM to manage Solidity dependencies. We should NPM install in these cases to ensure 
+      # Slither has the needed dependencies for scanning
+      if @repository.package_json_present?
+        run_shell("npm install", chdir: @repository.path_to_repo)
```
saikat056, line 20: It would be great if we can check whether the project is **hardhat** or **truffle**  before using `npm install`

ghbren:
Please add support for `--exclude-informational` and `--exclude-optimization`

ghbren:
Please also add support for `--filter-paths` in salus config.

ghbren:
being replaced with https://github.com/coinbase/salus/pull/623

PR #609, opened 2022-05-18
Fix for YarnAudit diff issue

maitrayshah:
- Updating YarnAuditSarif to handle diff issue when line numbers are mismatched.
- Updating nokogiri.


lib/sarif/yarn_audit_sarif.rb
```diff
@@ -68,5 +70,10 @@ def sarif_level(severity)
         SARIF_WARNINGS[:note]
       end
     end
+
+    def self.snippet_possibly_in_git_diff?(snippet, lines_added)
+      snippet += ":"
```
ghbren, line 75: Wouldn't this incorrectly match `aabc` and `abc`?

ghbren (approved):

PR #612, opened 2022-05-19
Adding check if dependencies key is present

maitrayshah:
NPMPackageVersionScanner

- Adding a check to validate `dependencies` are present in `package-lock.json`.
- Adding spec to handle the above use case.

NPMAudit

- Changing `should_run?` condition to include both `package.json` and `package-lock.json`
- Updating incorrect test case to handle above use case.


saikat056:
Can we add a unit test for this change? Thank you.

lib/salus/package_utils/package_lock_json.rb
```diff
@@ -56,9 +56,11 @@ def record_dep_locations
 
     # recursively store all dependency names as keys
     def get_dep_names(data)
-      data['dependencies'].each do |name, dep_info|
-        @deps[name] = {}
-        get_dep_names(dep_info) if dep_info['dependencies']
+      if data.key?("dependencies")
```
joshua-ostrom, line 59: return unless data.key?("dependencies")
might be closer to the ruby style guide 


Bigger ask though is can we add a spec for the behavior?

maitrayshah: Yup adding

lib/salus/package_utils/package_lock_json.rb
```diff
@@ -56,6 +56,8 @@ def record_dep_locations
 
     # recursively store all dependency names as keys
     def get_dep_names(data)
+      return unless data.key?("dependencies")
```
saikat056, line 59: It would be great if we can have a unit test related the changes in 'def get_dep_names'.

We can add the unit test in a separate pr.

It's not recommended to use get_* or set_* at the beginning of method name in ruby? Can we use a method name like 'def generate_dep_names_for(data)'.

We can make these changes in a separate pr.

saikat056 (approved):

PR #623, opened 2022-06-01
copy changes from slither spike and rewrite should_run?

ghbren:


lib/salus/scanners/slither.rb
```diff
@@ -0,0 +1,94 @@
...
+        shell_return = run_shell(pcmd, chdir: @repository.path_to_repo)
+        if !shell_return.success?
+          err_msg = pcmd + ' failed! ' + shell_return.stderr
+          report_error(err_msg)
+          report_stderr(err_msg)
+          return report_failure
+        end
+      end
+
+      slither_cmd = 'slither . --json - --exclude-informational --exclude-optimization'
```
swkanwar, line 34: Should we add a TODO comment to update this when we support configurability?

spec/fixtures/slither/no-compile-truffle/contracts/bad-contract.sol
```diff
@@ -0,0 +1,15 @@
+pragma solidity ^0.8.4;
+
+contract C {
+    function f() internal returns (uint a) {
```
swkanwar, line 4: Very minor comment feel free to ignore: Should we add a testcase for Ternary operator and assert that Slither fails? When we start supporting new version of slither(potentially with the fix), this testcase will fail and we will findout that the Ternary bug is fixed now

Dockerfile
```diff
@@ -68,6 +68,8 @@ RUN pip install wheel \
   && mv .local/bin/bandit .local/bin/bandit2 \
   && pip3 install --user bandit==${BANDIT_VERSION}
 
+ENV SLITHER_VERSION 0.8.3
+RUN pip3 install --user --no-cache-dir slither-analyzer==${SLITHER_VERSION}
```
saikat056, line 72: Can you please expand on the use of `--user` and `--no-cache-dir` options in the `pip3 install` command? I am curious to understand what their purposes are. 🤔

docs/scanners/slither.md
```diff
@@ -0,0 +1,8 @@
+# Slither Scanner
+
+Performs static analysis on solidity code.
+Only projects that use truffle or hardhat configs are supported.
+
+## Configuration
+
+Placeholder for future config options.
```
saikat056, line 8: We can add more details in this doc in the follow-up pr. Thank you for adding this placeholder doc.

lib/salus/scanners/slither.rb
```diff
@@ -0,0 +1,94 @@
...
+
+module Salus::Scanners
+  class Slither < Base
+    include Salus::Formatting
+
+    def should_run?
+      # only support truffle/hardhat for now
+      return false if !@repository.package_json_present? || !has_package_config
+
+      @repository.sol_file_present? != false
```
saikat056, line 15: Can we just write `@repository.sol_file_present?` as the last line of the method instead of using `@repository.sol_file_present? != false`. In other words, is the following code-snippet same?

```ruby
def should_run?
   ....
   @repository.sol_file_present?
end
```

lib/salus/scanners/slither.rb
```diff
@@ -0,0 +1,94 @@
...
+    end
+
+    def run
+      # Most Solidity projects will be Hardhat (https://hardhat.org/) or Truffle (https://trufflesuite.com/),
+      # which both use NPM to manage Solidity dependencies. We should NPM install in these cases
+      # to ensure Slither has the needed dependencies for scanning
+
+      pre_slither_commands = ['npm install', 'npm config set user 0']
```
saikat056, line 23: Is it possible to extract some parts of the code to separate methods. `def run` method becomes large. It's ok to have a number of small method instead of a long method. Martin Fowler has a great [article](https://martinfowler.com/bliki/FunctionLength.html) on this.

lib/salus/scanners/slither.rb
saikat056, line 34: What is the purpose of `--exclude-optimization` option? Can you please some details as comment? Thank you.

lib/salus/scanners/slither.rb
```diff
@@ -0,0 +1,94 @@
...
+      end
+
+      if !stdout_json['success'] || !stdout_json['error'].nil? ||
+          stdout_json['results'].nil? || stdout_json['results']['detectors'].nil?
+        err_msg = 'Error extracting slither output: ' + shell_return.inspect
+        report_error(err_msg)
+        report_stderr(err_msg)
+      else
+        results = []
+        ref_url_prefix = 'https://github.com/crytic/slither/wiki/Detector-Documentation#'
```
saikat056, line 56: Can we make this part as a constant?

REF_URL_PREFIX = 'https://github.com/crytic/slither/wiki/Detector-Documentation#'

spec/lib/salus/scanners/slither_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+    end
+
+    it 'should report error if solidity code has invalid syntax' do
+      repo = Salus::Repo.new('spec/fixtures/slither/no-compile-truffle')
+      scanner = Salus::Scanners::Slither.new(repository: repo, config: {})
+      scanner.run
+
+      expect(scanner.report.to_h.fetch(:passed)).to eq(false)
+      logs = scanner.report.to_h[:logs]
+      expect(logs).to eq(nil)
```
saikat056, line 128: `logs` variable has been used only once. So, we can avoid declaring this variable

`expect(scanner.report.to_h[:logs]) to be_nil`

spec/lib/salus/scanners/slither_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+        expect(scanner.version).to be_a_valid_version
+      end
+    end
+  end
+
+  describe '#supported_languages' do
+    context 'should return supported languages' do
+      it 'should return solidity' do
+        langs = Salus::Scanners::Slither.supported_languages
+        expect(langs).to eq(['solidity'])
```
saikat056, line 149: expect(described_class.supported_languages).to eq(['solidity'])

spec/lib/salus/scanners/slither_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+      expect(scanner.should_run?).to eq(false)
+    end
+
+    it 'should return false with .sol file, truffle config file but no package.json' do
+      repo = Salus::Repo.new('spec/fixtures/slither/solidity3')
+      scanner = Salus::Scanners::Slither.new(repository: repo, config: {})
+      expect(scanner.should_run?).to eq(false)
+    end
+
+    it 'should return true with .sol file, package.json, and truffle config' do
```
saikat056, line 39: Please split the example into multiple examples.

```ruby
it 'should return true with .sol file' do
  repo = Salus::Repo.new('spec/fixtures/slither/solidity_truffle_js')
  scanner = Salus::Scanners::Slither.new(repository: repo, config: {})
  expect(scanner.should_run?).to eq(true)
end

it 'should return true with package.json' do
end
```

spec/lib/salus/scanners/slither_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+      scanner = Salus::Scanners::Slither.new(repository: repo, config: {})
+      expect(scanner.should_run?).to eq(true)
+
+      # has package.json and truffle-config.ts
+      repo = Salus::Repo.new('spec/fixtures/slither/solidity_truffle_config_ts')
+      scanner = Salus::Scanners::Slither.new(repository: repo, config: {})
+      expect(scanner.should_run?).to eq(true)
+    end
+
+    it 'should return true with .sol file, package.json, and hardhat config' do
```
saikat056, line 61: Please split this example into multiple examples like this [comment](https://github.com/coinbase/salus/pull/623/files#r890308534).

saikat056:
@ghbren I left some some suggestions. We can address them in a separate PR if you want.

lib/salus/scanners/slither.rb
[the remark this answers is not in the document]
ghbren: We'll have a separate task to investigate.

lib/salus/scanners/slither.rb
ghbren: added comment.

ghbren: added comments

spec/lib/salus/scanners/slither_spec.rb
ghbren: updated accordingly

spec/lib/salus/scanners/slither_spec.rb
ghbren: updated

lib/salus/scanners/slither.rb
ghbren: Not really, because `@repository.sol_file_present?` does a wildcard search, which could return an array of file paths.

lib/salus/scanners/slither.rb
ghbren: done

spec/lib/salus/scanners/slither_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+
+    it 'should fail when there are vulnerabilities' do
+      repo = Salus::Repo.new('spec/fixtures/slither/solidity-bad')
+      scanner = Salus::Scanners::Slither.new(repository: repo, config: {})
+      scanner.run
+
+      expect(scanner.report.to_h.fetch(:passed)).to eq(false)
+      stdout = scanner.report.to_h[:info][:stdout]
+      stdout = JSON.parse(stdout)
+      expected_ref_url = "https://github.com/crytic/slither/wiki/Detector-Documentation#incorrect-shift"
```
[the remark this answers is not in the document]
ghbren: updated

lib/salus/scanners/slither.rb
ghbren: Updated

Dockerfile
ghbren: I copied this command from the spike.
The other `pip3` commands in Dockerfile also use `--user --no-cache-dir`.

`--no-cache-dir` seems to store some intermediate installation files.  According to this [stackoverflow post](https://stackoverflow.com/questions/45594707/what-is-pips-no-cache-dir-good-for), the cache is useless in a Docker image, and you can shrink the image size by with `--no-cache-dir`.

spec/lib/salus/scanners/slither_spec.rb
ghbren: done

spec/lib/salus/scanners/slither_spec.rb
ghbren: done

spec/fixtures/slither/no-compile-truffle/contracts/bad-contract.sol
ghbren: I subscribed to the slither notifications so I'll get notified when there is a new release.

swkanwar:
Changes Look good to me.

lib/salus/scanners/slither.rb
saikat056: Interesting. In this case, the method name should not end with '?'. It is convention in ruby to use '?' at the end of boolean method which returns true/false values. I will try to get a reference in ruby book later. But, this is an [article](https://medium.com/@sologoubalex/boolean-methods-in-ruby-94a2e907e5ea) on boolean method.

saikat056 (approved):

saikat056:
Approved and merged.

PR #631, opened 2022-06-09
[SECT-1483] Adding scanner_type for each scanner

swkanwar:
Adding support for scanner_type for each scanner

### DIFF
--- a/lib/salus/scanners/base.rb
+++ b/lib/salus/scanners/base.rb
@@ -53,6 +53,10 @@ module Salus::Scanners
       []
     end
 
+    def self.scanner_type
+      ''
+    end
+
     def version_valid?(version)
       return false if !version.is_a?(String)
 
--- a/lib/salus/scanners/base.rb
+++ b/lib/salus/scanners/base.rb
@@ -54,7 +54,7 @@ module Salus::Scanners
     end
 
     def self.scanner_type
-      ''
+      raise NoMethodError, 'implement in subclass'
     end
 
     def version_valid?(version)
--- a/lib/salus/scanners/brakeman.rb
+++ b/lib/salus/scanners/brakeman.rb
@@ -6,7 +6,6 @@ require 'tempfile'
 
 module Salus::Scanners
   class Brakeman < Base
-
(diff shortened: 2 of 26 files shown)


lib/salus/scanners/base.rb
```diff
@@ -53,6 +53,10 @@ def self.supported_languages
       []
     end
 
+    def self.scanner_type
```
saikat056, line 56:
```ruby
def self.scanner_type
  raise NoMethodError, 'implement in subclass'
end
```

Usually, in ruby, we force the developers to implement the method in subcalss by raising an exception. We have similar code in other [places](https://github.com/coinbase/salus/blob/9c93eb2e2b9d87b4a60de7bc1512c4f0556772ce/lib/salus/scanners/node_audit.rb#L143).

saikat056:
We can create a class of scanner_type constants and refer the constant in each scanner.

```ruby
module Salus::ScannerTypes
  class Types
       SAST = 'SAST'.freeze
       DAST = 'DAST'.freeze
  end
end
```

Then, refer to the constants from the scanner sub-class

```ruby
def self.scanner_type
   Salus::ScannerTypes::Types::SAST
end
```

lib/salus/scanners/brakeman.rb
```diff
@@ -6,6 +6,10 @@
 
 module Salus::Scanners
   class Brakeman < Base
+    def self.scanner_type
+      'SAST'
+    end
```
saikat056, line 11: If we have a class defining all [constants](https://github.com/coinbase/salus/pull/631#issuecomment-1151534600), then we can write this part as follows:

```ruby
def self.scanner_type
  Salus::ScannerTypes::Types::SAST
end
```

lib/salus/scanners/base.rb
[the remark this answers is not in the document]
swkanwar: Done, thank you!

swkanwar: Done

saikat056 (approved):

PR #633, opened 2022-06-10
Replaced Dir glob with ripgrep

spoken-angus:
Currently Salus uses a Dir search which is slow on large repositories
Replacing the Dir glob with ripgrep improves the performance by over 3x
Currently Salus also uses ripgrep in RepoSearcher


saikat056:
@spoken-angus Can you please provide details about this change in the description section? It would be great if you can talk about the **reason** for this change.

Please **DO NOT** share the link to the jira ticket or mention the ticket number. Thank you.

spoken-angus:
> @spoken-angus Can you please provide some details about this change in the description section? It would be great if you can talk about the **reason** for this change.
> 
> Please **DO NOT** share the link to the jira ticket here. Thank you. 😄

Added!

lib/salus/repo.rb
```diff
@@ -95,5 +95,18 @@ class Repo
     def initialize(path_to_repo = nil)
       @path_to_repo = path_to_repo
     end
+
+    ##
+    # @param [Array] args CLI rg command and options to run
+    # @return [Array<String>] Relative path of files returned
+    # from running the command.
+    def run_rg(*args)
```
saikat056, line 103: I am trying to understand. Which part of this file calls `run_rg`?

ghbren:
Could you add some notes on how this PR was tested?

lib/salus/repo.rb
```diff
@@ -55,7 +55,7 @@ class Repo
     IMPORTANT_FILES.each do |file|
       define_method :"#{file[:handle]}_present?" do
         if file[:wildcard]
-          files = Dir["#{@path_to_repo}/**/*#{file[:filename]}"]
+          files = run_rg("rg", "--files", "-g", file[:filename])
```
ghbren, line 58: What do `--files` and `-g` mean? could add a comment in the code explaining these?

lib/salus/repo.rb
```diff
@@ -95,5 +95,18 @@ class Repo
     def initialize(path_to_repo = nil)
       @path_to_repo = path_to_repo
     end
+
+    ##
+    # @param [Array] args CLI rg command and options to run
+    # @return [Array<String>] Relative path of files returned
+    # from running the command.
+    def run_rg(*args)
+      data = IO.popen(args, chdir: @path_to_repo).read
```
ghbren, line 104: I'm not seeing `chdir` as a valid option for `IO.popen`.  It is most likely has no effect in the call.
https://docs.ruby-lang.org/en/master/IO.html#method-c-popen

Please see the links under `Optional keyword arguments opts specify:`

lib/salus/repo.rb
```diff
@@ -95,5 +95,18 @@ class Repo
...
+      return [] if data == ""
```
ghbren, line 105: When would data be `""`?
Is it when a directory is empty?  

What about when data is `nil`?

lib/salus/repo.rb
ghbren: If. you want to run this from the right directory, you can do
```ruby
      data = ''
      Dir.chdir(path_to_repo) do
        data = IO.popen(args).read	
      end
```

lib/salus/repo.rb
ghbren: Maybe even add a brief description of `rg` with its url.

ghbren:
I'm seeing some code that is exactly the same as the code in repo_searcher.rb.
In that case, why not just call these functions in RepoSearcher?
We need to eliminate duplicated code when possible.

ghbren:
It seems like there is a bug with repo_searcher.rb.

spoken-angus:
> Was this PR tested outside the unit tests? Ideally we should docker build a local salus container and then docker run it.
> 
> Maybe just run it on one of our largest repos and make sure no weird issues, if you haven't already.

I ran it on mono/repo locally to test the PR

lib/salus/repo.rb
```diff
@@ -95,5 +95,18 @@ class Repo
...
+
+      files = data.lines.map { |file| File.join(path_to_repo, file.strip) }
```
saikat056, line 107: Do we need to create absolute path? Instead of this, can we use the relative path?

`files = data.lines.map(&:strip)`

spoken-angus: I was getting unit test errors when I used a relative path

lib/salus/repo.rb
```diff
@@ -95,5 +95,18 @@ class Repo
     def initialize(path_to_repo = nil)
       @path_to_repo = path_to_repo
     end
+
+    ##
+    # @param [Array] args CLI rg command and options to run
+    # @return [Array<String>] Relative path of files returned
```
saikat056, line 101: This comment says ` # @return [Array<String>] Relative path of files returned`. But, I think the method is returning absolute paths. Am I right? 🤔 

Can we please change the comment so that it is consistent with the method body?

saikat056:
Is it possible to show the comparison of run-time (ms) for **before** and **after** change? You can use ruby [benchmark](https://ruby-doc.org/stdlib-2.5.0/libdoc/benchmark/rdoc/Benchmark.html) to measure this time. Please run salus with this change for a very large project and display the difference.

Good work!! Thank you for this change.

spoken-angus:
> Is it possible to show the comparison of run-time (ms) for **before** and **after** change? You can use ruby [benchmark](https://ruby-doc.org/stdlib-2.5.0/libdoc/benchmark/rdoc/Benchmark.html) to measure this time. Please run salus with this change for a very large project and display the difference.
> 
> Good work!! Thank you for this change.

Performance Test Results:
Dir glob on mono/repo: 1.612924000000021 seconds
ripgrep on mono/repo 0.6968730000007781 seconds

2.31x performance increase

spoken-angus:
> I'm seeing some code that is exactly the same as in repo_searcher.rb. We should always prevent duplicated code.
> 
> Could you think about how to refactor the code to eliminate the duplicated code? For example, you can directly call `RepoSearcher#run_rg`, and then prepend `path_to_repo` to the results.
> 
> Slack me if you need help with this.

Just pushed the refactor

lib/salus/repo.rb
ghbren: Was the original code outputting absolute paths or relative paths?

lib/salus/repo.rb
```diff
@@ -51,14 +53,20 @@ class Repo
       { handle: :hardhat_config_ts, filename: 'hardhat.config.ts' }
     ].freeze
 
+    # Define config for RepoSearcher
+    config = { "pass_on_raise" => false, "scanner_timeout_s" => 0 }
```
ghbren, line 57: Why was this added?

spoken-angus: a config is required to initialize a `RepoSearcher`, this was the default

lib/salus/repo.rb
spoken-angus: It returned the absolute path

spoken-angus: Example of an error when using relative path
``` 
 3) Salus::Scanners::Bandit#should_run? setup.cfg present but no requirements.txt should return true
     Failure/Error: expect(repo.py_files_present?).to eq([py_file])

       expected: ["spec/fixtures/python/python_project_no_req_txt/main.py"]
            got: ["main.py"]
```

saikat056:
@spoken-angus Can you please rebase this PR? Looks like the pr is out of date with the base branch. Otherwise, the pr looks good to me. Thank you for the change.

saikat056 (approved):

saikat056 (approved):

PR #634, opened 2022-06-11
support slither --filter-paths

ghbren:
Support slither `--filter-paths`.

Ex. filter one path
```yaml
scanner_configs:
  Slither:
    filter-paths: contracts/bad-contract1.sol
```

Ex. filter multiple paths, separate by `|`
```yaml
scanner_configs:
  Slither:
    filter-paths: contracts/bad-contract1.sol|contracts/bad-contract2.sol
```

Tested with new unit tests and docker build/run local salus in contract repo.


geraldnash-cb:
Instead of separating a `|`-delimited string, can we make `filter-paths` a YAML array?

lib/salus/scanners/slither.rb
```diff
@@ -73,6 +74,17 @@ def self.supported_languages
 
     private
 
+    def config_opts
+      opts = ''
+      if @config['filter-paths'].to_s != ''
+        # Note: Slither webpage says filter-paths value is either a regular path or a regex
+        #       But `slither --help` says it is a comma separated list of paths, which is incorrect.
+        #       A list of paths must be joined by | as in a regular expression, not comma
+        opts += ' --filter-paths ' + @config['filter-paths'].to_s + ''
```
joshua-ostrom, line 83: Can we collect this as a yaml array and then apply the transformation to a | delimited list here?  That would keep our Yaml a bit more standard and give us a level of indirection as slither moves from | to , etc.

joshua-ostrom (approved):

PR #636, opened 2022-06-15
Add Slither Scanner Options to Ignore Informational and Optimization-Level Detectors

geraldnash-cb:


swkanwar (dismissed):

lib/salus/scanners/slither.rb
```diff
@@ -82,6 +82,12 @@ def config_opts
         #       A list of paths must be joined by | as in a regular expression, not comma
         opts += ' --filter-paths ' + @config['filter-paths'].to_s + ''
       end
+      if !@config.key?('exclude-informational') || @config['exclude-informational']
```
ghbren, line 85: can you use `dig` instead?

ghbren: Also, will your code behavior differently if
```yaml
exclude-information: "false"
```

vs

```yaml
exclude-information: false
```

geraldnash-cb: Yes `"false"` and `false` will cause different behavior since the first truthy and the second is falsey. Do we want to support multiple types for config options? I assumed we only wanted to accept boolean values for these options.

geraldnash-cb: Also, I think a `dig` call would work here if the condition were `@config.key?('exclude-informational') || @config['exclude-informational']`, but I think the fact that it's `!@config.key?('exclude-informational') || @config['exclude-informational']` (note the bang) makes this condition semantically different from `dig`.

lib/salus/scanners/slither.rb
```diff
@@ -82,6 +82,8 @@ def config_opts
         #       A list of paths must be joined by | as in a regular expression, not comma
         opts += ' --filter-paths ' + @config['filter-paths'].to_s + ''
       end
+      opts += ' --exclude-informational' if @config.dig('exclude-informational')
```
joshua-ostrom, line 85: If we see a need for additional options I would be a fan of moving this to a dynamic OPT list.  I'm okay w/the static listing given we just have  2 bool and 1 string opt but any more would be nice to DRY up.

saikat056 (approved):

PR #638, opened 2022-06-16
exit with new code if malformed exception found

ghbren:
Updated `report_error` on malformed exceptions to include `hard_error: true`, which will cause salus to fail with new exit status 2.

Otherwise, suppose 

1. there are vulnerabilities for a scanner
2. there is a malformed exception in salus.yaml for the scanner
3. `pass_on_raise=true`

Then the scan will pass and hide the vulnerabilities.

Tested with unit tests and docker build ... docker run.


lib/salus.rb
```diff
@@ -26,6 +26,7 @@ module Salus
 
   EXIT_SUCCESS = 0
   EXIT_FAILURE = 1
+  EXIT_HARD_ERROR = 2
```
joshua-ostrom, line 29: Can you provide a comment for the intended use.  This will ensure new devs can ramp up quickly on when they should leverage hard_error

joshua-ostrom: Any intent for non-configuration errors to be used here in the future?

ghbren: added comment

joshua-ostrom (dismissed):

joshua-ostrom (approved):

PR #646, opened 2022-06-23
Selecting resolved version in Gradle dependency graph

maitrayshah:
Currently when parsing the dependency gradle dependency graph there are instances when gradle [resolves to a different version](https://stackoverflow.com/questions/27952388/what-does-arrow-mean-in-gradles-dependency-graph), this change allows us to select the resolved version.

### DIFF
--- a/lib/salus/scanners/osv/gradle_osv.rb
+++ b/lib/salus/scanners/osv/gradle_osv.rb
@@ -76,6 +76,7 @@ module Salus::Scanners::OSV
         if dependency['version'].present?
           version = dependency['version']
           # Cleanup version string to handle case like -
+          # dependency.xx.xx:1.1.1 -> 1.5.0
           # 1.2.1.somestring / 9999.0-empty-to-avoid-conflict-with-test /
           # 30.3.0-deprecated-use-gradle-api
           version = version.split("->")[1].strip if version.include? "->"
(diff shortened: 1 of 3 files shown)


lib/salus/scanners/osv/gradle_osv.rb
```diff
@@ -76,8 +76,10 @@ def match_vulnerable_dependencies(dependencies)
         if dependency['version'].present?
           version = dependency['version']
           # Cleanup version string to handle case like -
-          # 1.2.1.somestring / 9999.0-empty-to-avoid-conflict-with-test /
-          # 30.3.0-deprecated-use-gradle-api
+          # Dynamic resolved version: dependency.xx.xx:1.1.1 -> 1.5.0
+          # Version strings: 1.2.1.somestring / 9999.0-empty-to-avoid-conflict-with-test /
+          # Version strings: 30.3.0-deprecated-use-gradle-api
+          version = version.split("->")[1].strip if version.include? "->"
```
saikat056, line 82: Can we use **last** instead of array_index in the case? Is the following code more readable?

`version.split("->").last.strip`

WDYT?

Here is a test

```
irb(main):001:0> s = "a->b"
=> "a->b"
irb(main):002:0> s.split('->')
=> ["a", "b"]
irb(main):003:0> s.split('->').last
=> "b"
```

saikat056 (approved):

lib/salus/scanners/osv/gradle_osv.rb
maitrayshah: Yup let me use that convention

saikat056 (dismissed):

saikat056:
@maitrayshah Do we have corresponding unit tests for this method? If not, can we a unit test in a separate pr?

PR #656, opened 2022-07-25
YarnAudit Auto Fix

maitrayshah:
Testing YarnAudit Auto Fix functionality.

### DIFF
--- a/lib/salus/scanners/yarn_audit.rb
+++ b/lib/salus/scanners/yarn_audit.rb
@@ -15,6 +15,7 @@ module Salus::Scanners
                   ' production --json'.freeze
     YARN_VERSION_COMMAND = 'yarn --version'.freeze
     BREAKING_VERSION = "2.0.0".freeze
+    YARN_INSTALL_COMMAND = 'yarn install'.freeze
 
     def should_run?
       @repository.yarn_lock_present?
@@ -119,6 +120,9 @@ module Salus::Scanners
 
       Salus::YarnLock.new(File.join(chdir, 'yarn.lock')).add_line_number(vulns)
       vulns = combine_vulns(vulns)
+
+      run_auto_fix(vulns)
+
       log(format_vulns(vulns))
       report_stdout(vulns.to_json)
       report_failure
@@ -184,6 +188,28 @@ module Salus::Scanners
       command << 'dependencies ' unless dep_types.include?('dependencies')
       command << 'devDependencies ' unless dep_types.include?('devDependencies')
       command << 'optionalDependencies ' unless dep_types.include?('optionalDependencies')
+    end
+
+    def run_auto_fix(vulnerabilities)
+      updates = {}
+      # This method forces indirect dependency updates
+      # There is an issue with package - lerna
+      # TODO - Validate if this works for direct dependency or it would need to
+      # updated directly in package.json
+      vulnerabilities.each do |vulnerability|
+        package = vulnerability["Package"]
+        regex = /^(#{package}.*?)\n\n/m
+        @repository.yarn_lock.gsub!(regex, '')
+      end
+
+      # Run yarn install to regenerate yarn.lock file
+      # Return contents of package.json and yarn.lock as results
+      reinstall_dependencies = run_shell(YARN_INSTALL_COMMAND)
+      if reinstall_dependencies.success?
+        updates["yarn.lock"] = @repository.yarn_lock
+        updates["package.json"] = @repository.package_json
+      end
+      updates
     end
 
     # severity and vuln title in the yarn output looks like


lib/salus/scanners/yarn_audit.rb
```diff
@@ -128,47 +126,172 @@ def handle_legacy_yarn_audit
...
-      package = vuln["Package"]
-      patched_version_range = vuln["Patched in"]
+    def fix_direct_dependency(feed)
+      @packages = JSON.parse(@repository.package_json)
+      feed.each do |vuln|
+        patch = vuln[:target]
+        resolves = vuln[:resolves]
+        package = vuln[:module]
+        resolves.each do |resolve|
+          if patch != "No patch available" && package == resolve[:path]
```
ghbren, line 178: is it possible for `patch` to be `nil` or empty string?

lib/salus/scanners/yarn_audit.rb
```diff
@@ -128,47 +126,172 @@ def handle_legacy_yarn_audit
...
+          end
+        end
+      end
+      write_auto_fix_files('package-autofixed.json', @packages.to_json)
+    end
 
-      if patched_version_range.match(Salus::SemanticVersion::SEMVER_RANGE_REGEX).nil?
-        report_error("Found unexpected: patched version range: #{patched_version_range}")
-        return
+    def fix_indirect_dependency(feed)
```
ghbren, line 186: It would safer to wrap the entire autofix part in a `begin ... rescue...`, just in case we encounter some thing weird.

lib/salus/scanners/yarn_audit.rb
```diff
@@ -128,47 +126,172 @@ def handle_legacy_yarn_audit
...
+      @parsed_yarn_lock = Salus::YarnLockfileFormatter.new(@repository.yarn_lock).format
+      subparent_to_package_mapping = []
+
+      feed.each do |vuln|
+        patch = vuln[:target]
+        resolves = vuln[:resolves]
+        package = vuln[:module]
```
ghbren, line 193: Are all these keys guaranteed to have values?

spec/lib/salus/scanners/yarn_audit_spec.rb
```diff
@@ -71,7 +71,7 @@
 
       expect(scanner.report.to_h.fetch(:passed)).to eq(false)
       vulns = JSON.parse(scanner.report.to_h[:info][:stdout]).sort { |a, b| a["ID"] <=> b["ID"] }
-      expect(vulns.size).to eq(17)
+      expect(vulns.size).to eq(18)
```
ghbren, line 74: there is a new vuln now?  was that added to the vuln DB recently? Just wanted to make sure the changes didn't break anything.

maitrayshah: I am probably guessing its a new vuln but will check it

lib/salus/scanners/yarn_audit.rb
maitrayshah: Since we are creating the feed object we should have the keys

lib/salus/yarn_formatter.rb
```diff
@@ -0,0 +1,40 @@
...
+      end
+      add_colon = ->(l) { l.sub(/(?<=\w|")\s(?=\w|")/, ": ") }
+
+      content.lines.map(&data).map(&add_colon).join
+    end
+
+    def parse_yaml_obj(yaml)
+      YAML.safe_load(yaml)
+    rescue StandardError
+      {}
```
ghbren, line 31: we should log an error here?

lib/salus/scanners/yarn_audit.rb
maitrayshah: I haven't seen an example of it but i can add a check

lib/salus/scanners/yarn_audit.rb
maitrayshah: Actually that might be better. Thanks

lib/salus/scanners/yarn_audit.rb
```diff
@@ -307,6 +503,14 @@ def deep_copy_wo_paths(vulns)
       vuln_list
     end
 
+    def is_major_bump(current, updated)
+      current_v = current.tr("^", "").tr("~", "").split('.').map(&:to_i)
```
ghbren, line 507: would this break with something like `~1.2.3-beta`?

maitrayshah: Yup let me update that, thinking of getting rid of all characters except . and number since in this case I only care about the major digit.

lib/salus/scanners/yarn_audit.rb
```diff
@@ -251,6 +386,16 @@ def scan_deps
       command << 'optionalDependencies ' unless dep_types.include?('optionalDependencies')
     end
 
+    def find_nested_hash_value(obj, key)
```
ghbren, line 389: It might be safer to just check if `obj.is_a?(Hash)`? 
Maybe some other random class also responds to `key?`

PR #662, opened 2022-08-08
Auto-Fix Yarn Audit Vulns in Direct Dependencies

geraldnash-cb:


lib/salus/scanners/yarn_audit.rb
```diff
@@ -127,6 +133,91 @@ def handle_legacy_yarn_audit
...
+      patched_version_range = vuln["Patched in"]
+
+      vulnerable_package_info = run_shell("yarn info #{package} --json").stdout
+      begin
+        vulnerable_package_info = JSON.parse(vulnerable_package_info)
+      rescue StandardError
+        report_error("yarn info #{package} --json did not return JSON")
+        return
+      end
+      list_of_versions = vulnerable_package_info["data"]["versions"]
```
ghbren, line 162: it would be safer to check if `vulnerable_package_info["data"]` is nil.
You can do something like below, not sure if my syntax is right
```ruby
vulnerable_package_info["data"].&["versions"]
```

ghbren:
Will yarn always output versions in the right format? like `1.2.3`?
Would it the version be something weird like `1A.2B.C`?

lib/salus/scanners/yarn_audit.rb
```diff
@@ -127,6 +133,91 @@ def handle_legacy_yarn_audit
       report_failure
     end
 
+    def auto_fix_vulns
```
ghbren, line 136: It would be safer to wrap the entire `auto_fix_vulns` in a `begin ... rescue ...`, just in case we get unexpected values.

lib/salus/scanners/yarn_audit.rb
```diff
@@ -127,6 +133,91 @@ def handle_legacy_yarn_audit
...
+        is_direct_dep = path == package
+
+        # TODO: Fix indirect dependencies as well
+        fix_direct_dependency(vuln) if is_direct_dep
+      end
+    end
+
+    def fix_direct_dependency(vuln)
+      package = vuln["Package"]
+      patched_version_range = vuln["Patched in"]
```
ghbren, line 153: Could `vuln["Patched in"]` be nil or empty string?

lib/salus/scanners/yarn_audit.rb
```diff
@@ -127,6 +133,91 @@ def handle_legacy_yarn_audit
...
+        if !@packages.dig("devDependencies", package).nil?
+          @packages["devDependencies"][package] = "^#{patched_version}"
+        end
+      end
+    end
+
+    def select_upgrade_version(patched_version_range, versions_list)
+      version_range_details =
+        patched_version_range
+          .match(/(?<operator>(<|>)?(=|~)?)?(?<major>\d+)\.(?<minor>\d+)\.(?<patch>\d+)/)
```
ghbren, line 183: Can `^` appear as an operator?

ghbren: Also, will yarn always output versions in the right format? like `1.2.3`?
Could it the version be something weird like `1A.2B.C`, or `1234`?

lib/salus/scanners/yarn_audit.rb
```diff
@@ -127,6 +133,91 @@ def handle_legacy_yarn_audit
...
+      begin
+        vulnerable_package_info = JSON.parse(vulnerable_package_info)
+      rescue StandardError
+        report_error("yarn info #{package} --json did not return JSON")
+        return
+      end
+      list_of_versions = vulnerable_package_info["data"]["versions"]
+      patched_version = select_upgrade_version(patched_version_range, list_of_versions)
+
+      if !patched_version.nil?
```
ghbren, line 165: The nested `if`s can cleaned up a little, like
```ruby
["dependencies", "resolutions", "devDependencies"].each  do |e|
  if ...
  end
end
```

lib/salus/scanners/yarn_audit.rb
```diff
@@ -31,6 +31,12 @@ def run
       else
         handle_legacy_yarn_audit
       end
+      if @repository.package_json_present?
+        @packages = JSON.parse(@repository.package_json)
+        # TODO: Serialize @packages to JSON and export the
+        #       updated contents to some package-autofixed.json file
+        auto_fix_vulns
```
ghbren, line 38: `auto_fix_vulns` should only be called if there are vulnerabilities.

lib/salus/scanners/yarn_audit.rb
geraldnash-cb: We can generally assume versions will follow `1.2.3` as you mentioned, because [NPM recommends semantic versioning](https://docs.npmjs.com/about-semantic-versioning) for packages. There may be edge cases that don't follow the standard, though.

lib/salus/scanners/yarn_audit.rb
geraldnash-cb: It could definitely be the latter.

lib/salus/scanners/yarn_audit.rb
```diff
@@ -127,6 +135,100 @@ def handle_legacy_yarn_audit
...
+      end
+    rescue StandardError
+      report_error("An error occurred while auto-fixing vulnerabilities")
+    end
+
+    def fix_direct_dependency(vuln)
+      package = vuln["Package"]
+      patched_version_range = vuln["Patched in"]
+
+      puts patched_version_range.match(SEMVER_RANGE_REGEX)
```
ghbren, line 159: I think you forgot to remove this debugging statement?

lib/salus/scanners/yarn_audit.rb
```diff
@@ -127,6 +135,100 @@ def handle_legacy_yarn_audit
...
+      return if patched_version_range.match(SEMVER_RANGE_REGEX).nil?
```
ghbren, line 160: do we also want to report error here?

ghbren (approved):

PR #717, opened 2022-10-06
fix semgrep non-uniq ids for forbidden patterns

ghbren:
In the current master, every semgrep forbidden pattern ends with ruleID `Forbidden Pattern Found` in the sarif.  

This PR updates the ruleID to be the `id` specified for the pattern in the semgrep config.
In case the pattern is directly specified without id in salus.yaml and a semgrep config is not used, then the `id` would be set to the sha1 hash of the pattern.

### DIFF
--- a/lib/salus/scanners/semgrep.rb
+++ b/lib/salus/scanners/semgrep.rb
@@ -1,5 +1,6 @@
 require "salus/scanners/base"
 require "json"
+require 'digest/sha1'
 
 # Report any matches to a list of semantic grep patterns provided by the config file.
 # semgrep is a grep like tool for doing AST pattern matching.
@@ -98,6 +99,7 @@ module Salus::Scanners
             if match["required"]
               failure_messages << "\nRequired #{user_message} was not found " \
               "- #{match['message']}"
+
               all_misses << {
                 pattern: match['pattern'],
                 config: match['config'],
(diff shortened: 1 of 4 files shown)


lib/salus/scanners/semgrep.rb
```diff
@@ -114,7 +115,13 @@ def run
                 "- #{msg}\n" \
                 "\t#{hit_to_string(hit, base_path)}"
               end
+              hit_id = if hit['check_id'] == '-' # id not specified by user
```
maitrayshah, line 118: `check_id` is always present right ?

ghbren: No.  

If you directly use a semgrep config, then the `check_id` is always there in the semgrep config.
```yaml
# salus.yaml
scanner_configs:
  Semgrep:
    matches:
      - config: semgrep_config_1.yaml
        forbidden: true
```

```yaml
rules:
  - id: eqeq-always-true
    patterns:
      - pattern: $X == $X
      - pattern-not: 0 == 0
    message: "$X == $X is always true"
    languages: [python]
    severity: ERROR
```

But, you can also directly specify the pattern in the salus.yaml, and there would be no ids.  Like
```yaml
scanner_configs:
  Semgrep:
    matches:
      - pattern: $X == $X
        message: Useless equlity check
        language: python
        forbidden: true
```

maitrayshah: Was just wondering if we want to enforce ids but i am okay with using if present if not hashing it.

maitrayshah (approved):

PR #737, opened 2022-10-25
Updating Report Node Modules Scanner

maitrayshah:
- Updating command and parsing entire output rather than only `--name-only` which results in a cleaner data.


lib/salus/scanners/report_node_modules.rb
```diff
@@ -102,22 +102,17 @@ def record_dependencies_from_latest_yarn_lock
...
+        line = JSON.parse(line)
+        name = line.dig("value")
+        version = line.dig("children", "Version")
+
+        unless name.nil? && version.nil?
           dependencies.append(
             {
-              "name" => name.delete('"').strip,
-              "version" => version.delete(':"').strip
+              "name" => name.split("@npm:")[0],
```
saikat056, line 114: `"name" => name.split("@npm:").first,`

It is more readable. WDYT? 🤔

saikat056 (approved):

saikat056 (approved):

PR #738, opened 2022-10-26
Invoke a particular scanner after all the report scanners are done execution

swkanwar:
Adding business logic to Invoke a particular scanner after all the report scanners are done execution
Updated processor.rb to create a list of reporting threads and added a Join step to wait for completion of these threads
In the Scanner which needs a event to unblock we need to add code like the following

```
$reporting_scanners_running = true
    def self.reporting_scanners_ran(event_name, data)
      # For now call to this fundtion means all reporting scanners are done
      $reporting_scanners_running = false
    end

```

In the run method add a blocking call 

```
 while $reporting_scanners_running do
        puts "Swaraj bundleaudit in while loop"
        sleep(1)
      end

      puts "Swaraj BundleAudit after the while loop has finished"
```
Register the listener
```
Salus::PluginManager.register_listener(Salus::Scanners::BundleAudit)

```


lib/salus/scanners/base.rb
```diff
@@ -53,6 +53,12 @@ def self.supported_languages
       []
     end
 
+    # Must be overriden by report scanners and should return true.
+    # This will be used to unblock other scanners which are dependent on Reporting Scanners output
+    def is_reporting_scanner
```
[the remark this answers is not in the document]
swkanwar: Done 👍

lib/salus/scanners/base.rb
```diff
@@ -53,6 +53,12 @@ def self.supported_languages
       []
     end
 
+    # Must be overriden by report scanners and should return true.
+    # This will be used to unblock other scanners which are dependent on Reporting Scanners output
+    def is_reporting_scanner?
```
joshuaostrom, line 58: If we find additional reporting feature needs we could have our scanner subclass a common report scanner class and then simply use .kind_of?(ReportScanner) here and remove the need for each class to define a is_reporting_scanner?

lib/salus/processor.rb
```diff
@@ -96,13 +96,14 @@ def fetch_config_file(source_uri, repo_path)
...
       threads = []
+      reporting_threads = []
       files_copied = []
       reraise_exceptions = ENV.key?('RUNNING_SALUS_TESTS')
 
       # We won't auto cleanup - we'll manually delete any copied fields after
       # the various threads have finished
       copied = RepoSearcher.new(@repo_path, config, false).matching_repos do |repo|
-        threads << Thread.new do
+        thread = Thread.new do
```
joshuaostrom, line 106: This works.  I could see a case where threads becoming scanning threads but as written what you have will work.
```
if canner_instance.is_reporting_scanner?
  reporting_threads << thread
else
  scanning_threads << thread
end
```

lib/salus/processor.rb
```diff
@@ -148,12 +153,15 @@ def scan_project
...
-
+          scanner_threads, copied, scanner_reporting_threads =
+            run_scanner(config, scanner_class, scanner_name)
           threads.concat(scanner_threads)
+          reporting_threads.concat(scanner_reporting_threads)
           files_copied.concat(copied)
         end
 
+        reporting_threads.each(&:join)
+        Salus::PluginManager.send_event(:reporting_scanners_ran, @scanners_ran, @report)
```
joshuaostrom, line 164: We should probably add a spec to test this new behavior

lib/salus/scanners/base.rb
ghbren: I looked at a bunch of `Report...` scanners, and all have the function defined below, and it doesn't look like `SBOM_REPORT` is used in any non-`Report...` scanners.  Perhaps just use that info to determine if the scanner is a reporting scanner?
Though you may want to double check if each of the report scanners has `SBOM_REPORT` as its type.

```
    def self.scanner_type
      Salus::ScannerTypes::SBOM_REPORT
    end
```

swkanwar: @joshuaostrom Scanner_type also seems to be a good way of knowing if its a reporting scanner, I can pivot to using this as well

joshuaostrom: That works for me!! Thanks all!

lib/salus/processor.rb
swkanwar: Done

lib/salus/scanners/base.rb
swkanwar: Done

lib/salus/scanners/report_base.rb
```diff
@@ -0,0 +1,8 @@
+module Salus::Scanners
+  # Super class for all scanner objects.
+  class ReportBase < Base
```
ghbren, line 3: Is there something that ensures this file is loaded before all classes that inherit from it?
Maybe by chance, it would work ok, but files may be loaded in random orders.
Perhaps it would be best to explicitly `require` this file first.

swkanwar: Good point, Done and added explicit "requires"

ghbren (approved):

PR #742, opened 2022-11-13
Fix Enforced Scanner Bug

maitrayshah:
Currently, certain scanners are treated as enforced even when they are not specified as enforced because of how default config is merged with user config. This PR changes the default config to not make any scanner as enforced.


saikat056 (dismissed):

saikat056:
@maitrayshah This pr will need a rebase

maitrayshah:
Gonna hold off on this one, since this changed default behavior.

PR #743, opened 2022-11-13
ReportNodeModules (yarn1.x) - Add Distribution attribute to SBOM

maitrayshah:


maitrayshah:
@saikat056 Added a sample to output distribution url

saikat056:
Looks good to me. Can we add some unit tests for this change? Thank you.

maitrayshah:
Oh yeah I mean this is not 100% complete but just an example of how we can populate distribution urls, were you able to generate a SARIF and check results and if they were what you were looking for ?

PR #746, opened 2022-11-20
Add salusWarnMessage flag

maitrayshah:
Adding `salusWarn` flag to SARIF to allow differentiating between warn and enforced scanner.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -148,6 +148,7 @@ enforced_scanners:
 # Defines configuration relevant to specific scanners.
 scanner_configs:
   BundleAudit:
+    warn_message: true # Used to update salusWarnMessage field in SARIF report. This attribute allows for more customization further downstream in the CI pipelines.
     ignore:
       - CVE-XXXX-YYYY # irrelevant CVE which does not have a patch yet
     recursion: # optional recusion settings.  
@@ -174,7 +175,6 @@ scanner_configs:
       # copied to sub directories for proper scanning.
         - Gemfile
         - Gemfile.lock
-      warn_message: true # Used to update salusWarnMessage field in SARIF report. This attribute allows for more customization further downstream in the CI pipelines.
 ```
 
 Special configuration that exist for particular scanners is defined in the [scanners directory](/docs/scanners).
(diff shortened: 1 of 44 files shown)


lib/sarif/base_sarif.rb
```diff
@@ -25,12 +25,21 @@ def initialize(scan_report, config = {}, repo_path = nil)
...
     def base_path
       @base_path ||= @repo_path.nil? ? nil : File.expand_path(@repo_path)
     end
 
+    def check_warn
+      unless @scanner_config.nil?
+        @scanner_config.dig(:scanner_configs, @scan_report.scanner_name, "warn_message") || false
+      end
+
+      false
```
ghbren, line 40: Doesn't this method always return `false`?

lib/sarif/cargo_audit_sarif.rb
```diff
@@ -4,10 +4,11 @@ class CargoAuditSarif < BaseSarif
 
     CARGO_AUDIT_URI = 'https://github.com/RustSec/cargo-audit/'.freeze
 
-    def initialize(scan_report, repo_path = nil)
+    def initialize(scan_report, repo_path = nil, scanner_config = {})
       super(scan_report, {}, repo_path)
       @uri = CARGO_AUDIT_URI
       @logs = parse_scan_report!
+      @scanner_config = scanner_config
```
ghbren, line 11:
```ruby
@scanner_config = scanner_config
``` 
appears in multiple sarif adapters.  Is there a way to set it in `super()` instead?
If I'm not mistaken, `super()` is called from every adapter.

lib/sarif/base_sarif.rb
maitrayshah: Fixed, thanks for catching that

lib/sarif/cargo_audit_sarif.rb
maitrayshah: When generating SARIF reports, it invokes the scanner sarif adapter which then populates the base sarif.

docs/configuration.md
```diff
@@ -174,6 +174,7 @@ scanner_configs:
       # copied to sub directories for proper scanning.
         - Gemfile
         - Gemfile.lock
+      warn_message: true # Used to update salusWarnMessage field in SARIF report. This attribute allows for more customization further downstream in the CI pipelines.
```
ghbren, line 177: Seems like this is not intended correctly? It's nested in `recursion`.

ghbren (approved):

PR #762, opened 2022-12-06
support for show_syntax_errors semgrep config option

ghbren:


lib/salus/scanners/semgrep.rb
```diff
@@ -89,6 +90,11 @@ def run
           data = JSON.parse(shell_return.stdout)
           hits = data["results"]
           semgrep_non_fatal_errors = data["errors"]
+          if @config['show_syntax_errors'].to_s == 'false'
```
maitrayshah, line 93: Do we need string comparison here, shouldn't it work without it ?

ghbren: Great catch! removed the `to_s`.

maitrayshah (approved):

PR #786, opened 2023-01-06
Granular thread grouping

joshuaostrom:
Updating the recent Salus event `reporting_scanners_ran` to provide a more granular set of integrations via a new `scanning_group_completed` event.

This event will allow plugins to detect when any of the following scanner types have completed scanning:
Salus::ScannerTypes::SBOM_REPORT, Salus::ScannerTypes::LICENSE, Salus::ScannerTypes::DEPENDENCY, Salus::ScannerTypes::SAST, Salus::ScannerTypes::DYNAMIC

If a particular Scanner Class want's to opt-out of the grouping that class can simply define a block_scanner_group? class method that returns false (default is true)

The prior event was undocumented, took the liberty to refactor the event signature.  Given that the event was not yet documented this should be acceptable but we will want to make sure the backwards incompatibility is reflected with the next version bump.


swkanwar (dismissed):
Looks great!

swkanwar (approved):

PR #830, opened 2023-03-28
Trufflehog - Exclude files

maitrayshah:
Support excluding files for Trufflehog Scanner.


joshua-ostrom:
What's the backstory?  When are users needing to allow artifacts with valid credentials?

